PR #25, opened 2020-05-27
Tickets/dm 25063

teweitsai:
Please ignore the commit: 6c6cedd in the review. Thanks!


setup.cfg
```diff
@@ -1,7 +1,7 @@
 [flake8]
 max-line-length = 110
 max-doc-length = 79
-ignore = E133, E226, E228, N802, N803, N806, N812, N813, N815, N816, W504
+ignore = E133, E203, E226, E228, N802, N803, N806, N812, N813, N815, N816, W503, W504
```
pkubanek, line 4: Not sure why we have to ignore E203 (space before :) and W503. flake8 runs fine without it in ignore list. I will rather keep the old setup.

teweitsai: The format of `black` will break the PEP-8 rules of E203 and W503. Therefore, we need to ignore them. Otherwise, the `scons` will fail.
PS. I used the default setting of `black` to reformat the code. Although we could change the default setting of `black` to follow this (I am not sure how to do this yet), I feel it is the easier just to use the default setting. Or we need to document the setting we use for `black`. I think this is annoying.

PR #28, opened 2020-09-15
Tickets/dm 26502

teweitsai:


python/lsst/ts/ofc/ctrlIntf/OFCCalculationFactory.py
```diff
@@ -51,7 +52,7 @@ def getCalculator(instName, state0Dof=None):
         Returns
         -------
         OFCCalculationOfLsst, OFCCalculationOfComCam, OFCCalculationOfSh, or
-        OFCCalculationOfCmos
+        OFCCalculationOfCmos, OFCCalculationOfLsstFam
```
pkubanek, line 55: misplaced or - shall be before OFCCalculationOfLsstFam, not before OFCCalculationOfCmos. Consider replacing the list with OFCCalculation child (OFCCalculationOfComCam).

I am not sure, but my understanding is that you shall drop OFCCalculationOfLsst class, as it is now (renamed?) OFCCalculationOfLsstFam - at least drop of LSST from InstName suggest this.

python/lsst/ts/ofc/ctrlIntf/OFCCalculationOfLsstFam.py
```diff
@@ -0,0 +1,44 @@
...
+        Parameters
+        ----------
+        state0Dof : dict, optional
+            State 0 DoF dictionary. If None (=default), the instrument's
+            default will be used. See
+            :lsst:ts:ofc:`OptCtrlDataDecorator.getState0FromDict` for format
+            details.
+        """
+
+        super(OFCCalculationOfLsstFam, self).__init__(
```
pkubanek, line 42: Python3 comes with slightly shorter version of super: super().__init__(, name of the class not needed.

pkubanek (changes requested):
Please see Python3 super() (easier than in Python2) usage.

python/lsst/ts/ofc/ctrlIntf/OFCCalculationFactory.py
teweitsai: Fixed in the commit: c90538b. Thanks!

python/lsst/ts/ofc/ctrlIntf/OFCCalculationOfLsstFam.py
teweitsai: Fixed in the commit: c90538b. Thanks!

pkubanek (approved):

PR #31, opened 2021-04-12
Tickets/dm 29687

tribeiro:
Small update to the CI job to build and upload documentation to ts-ofc.lsst.io.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,9 +8,13 @@ pipeline {
         // It is recommended by SQUARE team do not add the label and let the
         // system decide.
         docker {
-            image 'lsstsqre/centos:w_latest'
-            args '-u root'
+          image 'lsstts/develop-env:develop'
+          args "-u root --entrypoint=''"
         }
+    }
+
+    options {
+      disableConcurrentBuilds()
     }
 
     triggers {
@@ -20,12 +24,19 @@ pipeline {
     environment {
         // Position of LSST stack directory
         LSST_STACK = "/opt/lsst/software/stack"
+        // PlantUML url
+        PLANTUML_URL = "https://managedway.dl.sourceforge.net/project/plantuml/plantuml.jar"
         // Pipeline stack Version
         STACK_VERSION = "current"
         // XML report path
         XML_REPORT = "jenkinsReport/report.xml"
         // Module name used in the pytest coverage analysis
         MODULE_NAME = "lsst.ts.ofc"
+        // Authority to publish the document online
+        user_ci = credentials('lsst-io')
+        LTD_USERNAME = "${user_ci_USR}"
+        LTD_PASSWORD = "${user_ci_PSW}"
+        DOCUMENT_NAME = "ts-ofc"
     }
 
     stages {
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -112,6 +110,47 @@ pipeline {
...
+                  pip install ltd-conveyor
+
+                  ltd upload --product ${env.DOCUMENT_NAME} --git-ref \${GIT_BRANCH} --dir doc/_build/html
+                    """
+
+                if ( RESULT != 0 ) {
+                    unstable("Failed to push documentation.")
+                }
+              }
+            }
```
teweitsai, line 143: Why you do not want to do this step in the **stages** but in the "*post**? I felt it might be cleaner to do it in the **stages**.

doc/conf.py
```diff
@@ -3,19 +3,20 @@
 This configuration only affects single-package Sphinx documentation builds.
 """
 
-from documenteer.sphinxconfig.stackconf import build_package_configs
-import lsst.ts.ofc
+import os.path
+from documenteer.conf.pipelinespkg import *  # noqa
```
teweitsai, line 7: I think you could put `# noqa` to `setup.cfg` instead to ignore the checking?

doc/conf.py
```diff
@@ -3,19 +3,20 @@
...
+import lsst.ts.ofc  # noqa
 
+project = "ts_ofc"
+html_theme_options["logotext"] = project  # noqa
```
teweitsai, line 11: I think you could put # noqa to setup.cfg instead to ignore the checking?

doc/conf.py
```diff
@@ -3,19 +3,20 @@
 This configuration only affects single-package Sphinx documentation builds.
 """
 
-from documenteer.sphinxconfig.stackconf import build_package_configs
-import lsst.ts.ofc
+import os.path
+from documenteer.conf.pipelinespkg import *  # noqa
+import lsst.ts.ofc  # noqa
```
teweitsai, line 8: I think you could put # noqa to setup.cfg instead to ignore the checking?

Jenkinsfile
```diff
@@ -8,8 +8,8 @@ pipeline {
         // It is recommended by SQUARE team do not add the label and let the
         // system decide.
         docker {
-            image 'lsstsqre/centos:w_latest'
-            args '-u root'
+          image 'lsstts/develop-env:develop'
```
teweitsai, line 11: Why do you want to use `lsstts/develop-env:develop` instead of `lsstsqre/centos:w_latest`? The package manager in `ts_ofc` is `eups` instead of `conda` now.

doc/index.rst
```diff
@@ -28,6 +28,12 @@ Important classes:
 * `BendModeToForce` transforms the bending mode to actuator forces and vice versa.
 * `SubSysAdap` transforms the degree of freedom or actuator forces from ZEMAX coordinate to subsystem's coordinate and vice versa.
 
+.. uml:: uml/ctrlIntfClass.uml
+    :caption: OFCCalculation class diagram.
```
teweitsai, line 32: This class diagram is more like the package of `ctrlIntf` instead of `OFCCalculation` itself.

doc/index.rst
```diff
@@ -28,6 +28,12 @@ Important classes:
...
+
+.. uml:: uml/ofcClass.uml
+    :caption: ZTAAC class diagram.
```
teweitsai, line 35: This class diagram is more like the package of `OFC` instead of `ZTAAC` itself.

Jenkinsfile
teweitsai: I saw the following in the history:
```
Use develop-env image for the CI job, due to the need of java to build the documentation.
```

It looks like the lsst image does not support this anymore. So sad...

teweitsai (approved):
This PR looks good to me. I have some comments maybe you could consider before the merge. Thanks!

Jenkinsfile
tribeiro: yeah... I think the problem is that, if I do it in a stage and a stage before it fails, this will not run. Also, if this is a stage and it fail, the stages bellow it will not run.

PR #32, opened 2021-04-21
DM-29730: Refactor ofc code

tribeiro:
* The code was reorganized so that the main user-interface class (previously OFCCalculation, now just OFC) is at the top level of the package.
* The OFC class is no longer subclassed for the different instruments, since the behavior was the same in all cases.

Data handling
-------------------

All data is now handled by a single container class, OFCData, which is shared between all the other classes that require data access.
Most of the data is defined in-line, instead of reading them from files.
The data that is read from files are the ones for the instruments.
This is handled in the background when the user sets the "name" attribute in an instance of OFCData.
All data is read at once and stored in memory to avoid unnecessary IO during computation.
In the future we may consider adding more data protection and data parsing capabilities, but I think this will work fine for this first iteration.

Corrections
---------------

Corrections is now handled by a single class.

Removed WEP dependency
-------------------------------------

OFC now does not depend on WEP anymore.
The dependency was mostly on enumerations for filter names and other things that mapped to string/filenames.
These where all replaced by true strings.
The advantage is that OFC is no longer tied to that particular use cases, if the user can provide configuration files that matches the input they can use it with WEP without any code changes.

Controller and State Estimator
-----------------------------------------

The core classes where previously called `OptCtrl`  and `OptStateEsti`.
They where renamed `OFCController` and `StateEstimator`.


teweitsai (approved):
This update looks good to me. I have some suggestions that will send in slack. Please consider them before the merge. Thanks!

teweitsai:
Please help to update the `README.md`. Thanks!

PR #41, opened 2021-07-15
Tickets/dm 31078

teweitsai:


python/lsst/ts/ofc/ofc.py
```diff
@@ -251,7 +251,7 @@ def get_all_corrections(self):
         return corrections
 
     def get_correction(self, dof_comp):
-        """Get the hexapod correction.
+        """Get the correction.
```
tribeiro, line 254: how about?

> Get correction for specified component.

tests/test_ofc.py
```diff
@@ -127,20 +127,7 @@ def test_set_pssn_gain(self):
...
-
-        (
-            m2_hex_corr,
-            cam_hex_corr,
-            m1m3_corr,
-            m2_corr,
-        ) = self.ofc.calculate_corrections(
-            wfe=wfe, field_idx=field_idx, filter_name=filter_name, gain=gain, rot=rot
-        )
+        m2_hex_corr, cam_hex_corr, m1m3_corr, m2_corr = self._calculate_corrections()
```
tribeiro, line 130: I am just going to point out that this method is called only once in the entire code base.. I honestly don't think it is worth breaking it down this way, it mades it harder to understand what is happening here, not easier...

tests/test_ofc.py
```diff
@@ -344,6 +343,29 @@ def test_truncate_dof(self):
...
+
+        # Second time of calculation
+        # The DOF should be aggregated
+        correction = self._calculate_m2_hex_correction()
+
+        self.assertAlmostEqual(correction[0], 5.2979926)
+        self.assertAlmostEqual(correction[1], -2.5117287)
+        self.assertAlmostEqual(correction[2], 19.6478187)
+
+    def _calculate_m2_hex_correction(self):
```
tribeiro, line 363: This is another case where I would strongly advocate for not having a private method. These are basically 3 lines of code that are way more comprehensive to have inline in the test method above than having on a private method.

doc/versionHistory.rst
```diff
@@ -6,6 +6,13 @@
 Version History
 ##################
 
+.. _lsst.ts.ofc-2.0.2:
+
+v2.0.2
+======
+
+* ``OFC.get_correction()`` should give the aggregated degree of freedom (DOF).
+
```
tribeiro, line 15: Can you please add an item to indicate that you removed out of date documents?

tribeiro (approved):
Looks good... I left some minor comments I hope you will look before merging.

python/lsst/ts/ofc/ofc.py
bxin: How about also indicating whether this is aggregated correction or just what is determined from the last visit?

teweitsai: Done in the commit: e64d8ae. Thanks!

tests/test_ofc.py
teweitsai: This `self._calculate_corrections()` was called more than once actually. But I removed this and put it back if this is what you want. The update is in commit: e64d8ae. Thanks!

tests/test_ofc.py
teweitsai: After I removed the `self._calculate_corrections()`, this function is 7 lines of code now. Therefore, I will keep it to reduce the duplication. Thanks!

doc/versionHistory.rst
teweitsai: Updated in the commit: e64d8ae. Thanks!

PR #42, opened 2021-07-22
Fix `setup.cfg` to support conda versioning.

tribeiro:


mareuter (approved):
That's the right thing.

PR #48, opened 2022-11-07
Update Jenkinsfile to post  to aos-builds 

suberlak:


Jenkinsfile
```diff
@@ -23,7 +23,7 @@ pipeline {
         // saluser home directory
         SALUSER_HOME = "/home/saluser"
         // PlantUML url
-        PLANTUML_URL = "https://github.com/plantuml/plantuml/releases/download/v1.2022.1/plantuml-1.2022.1.jar"
```
teweitsai, line 26: Why not from GitHub? https://github.com/plantuml/plantuml/releases

teweitsai:
Please help to update the version history. Thanks!

Jenkinsfile
suberlak: I was using the one from `ts_phosim` https://github.com/lsst-ts/ts_phosim/blob/2a06807d495eb21e38aac196b0c88d6842de0f72/Jenkinsfile#L35   but I see that this seems to be the cause in the recent `ts_phosim` build https://tssw-ci.lsst.org/job/LSST_Telescope-and-Site/job/ts_phosim/job/develop/593/execution/node/67/log/  - should we update the `ts_phosim` one to `https://github.com/plantuml/plantuml/releases/download/v1.2022.12/plantuml.jar` as well ?

suberlak: While I updated Jenkinsfile I asked Tiago if we want to keep the hard-coded version  `https://github.com/plantuml/plantuml/releases/download/v1.2022.1/plantuml-1.2022.1.jar` or link  to most recent stable release  `http://sourceforge.net/projects/plantuml/files/plantuml.jar`, and the latter seemed a better choice.  If you think it's better to use the github one I'll revert that commit.

teweitsai: I thought we used the `sourceforge.net` in the initial beginning. But it was so unstable that we changed to use the `GitHub` in the final. For example:

https://github.com/lsst-ts/ts_wep/blob/develop/Jenkinsfile#L36

BTW, I think the PlantUML is very stable now and the version's difference is not much.

teweitsai: History:
https://github.com/lsst-ts/ts_ofc/commit/b5f2e009f631da08ac88b2c333030be090d4d57e

suberlak: Reverted to the original (github) version

PR #50, opened 2023-07-18
Tickets/DM-40085

gmegh:


suberlak (approved):
This is a slightly different solution than https://github.com/lsst/meas_algorithms/pull/335/files - does it work with Python3.11 ?

suberlak:
If it does work with lsst-scipipe-7.0.1 , then it's just a question of DM style recommendations.

PR #51, opened 2023-07-20
Tickets/DM-39910

gmegh:
This PR addresses the rotation issues in ts_ofc along with various other issues that needed to be fixed. See below for a detailed list of the changes introduced:

- **Rotation:**
     -  Intrinsic zernikes are now a Double Zernike object. This allows us to find the right intrinsic coefficients when the camera is rotated, by evaluating the intrinsic Zernike object at the sensors, after rotating it. 
     -  Sensitivity matrix is now a Double Zernike object. This allows us to find the right sensitivity matrix when the camera is rotated. We first rotate the double Zernike object to match the field rotation and then find the values at the sensors of interest.
     - Added two scripts to generate intrinsic and sensitivity matrix double Zernike objects with batoid simulation package.
     - The estimated Zernikes coming from ts_wep are de-rotated prior to any calculation to take into account the sensor rotation due to the camera rotation. 
- Added a `remove_degeneracies` function that removes degenerated modes from the computed corrections based on a threshold for the singular value set as `self.ofc.rcond_degeneracy`
- Added unit tests for the Double Zernike Sensitivity matrix
- Refactored all unit tests. Removed hardcoded values and generalized the corrections check, to be able to use any initial state or wavefront error. Having hardcoded values was an issue when changing the values of the sensitivity matrix. We expect the sensitivity matrix to change in the future, so the unit tests must not depend on prior sensitivity matrix values. 
- Minor formatting fixes of the code. Adding expected input types and output types. Adding comments to functions.
- Removed CamRot class, as it is no longer needed to derotate corrections. All the rotation is handled at the beginning.
- Removed unused test data from testData folder. 

_Bonus:_
For your reassurance, two plots are attached below comparing the old sensitivity matrix and intrinsic Zernike values to the updated ones. We agree up to 0.1% for the sensitivity matrix and up to ~5% for the intrinsic zernikes, except for the u band. 

Maximum difference for each degree of freedom between the new Double Zernike Sensitivity matrix evaluated at the 35 sample points and the previous ts_ofc sensitivity matrix. 

Screenshot 2023-08-21 at 13 37 14 (1)

Difference between the new Double Zernike Intrinsic Zernikes evaluated at the 35 sample points and the previous ts_ofc intrinsic zernikes. All the filters are in agreement, except for 'u', for which the previous ts_ofc values were off.
 
Screenshot 2023-08-21 at 10 02 05


python/lsst/ts/ofc/sensitivity_matrix.py
```diff
@@ -0,0 +1,152 @@
...
+                np.array(
+                    [
+                        zk.coef
+                        for zk in galsim.zernike.DoubleZernike(
+                            self.ofc_data.sensitivity_matrix[:, :, dof_idx],
+                            # Rubin annuli
+                            uv_inner=0.0,
+                            uv_outer=1.75,
+                            xy_inner=0.612 * 4.18,
+                            xy_outer=4.18,
```
teweitsai, line 106: These parameters should be in the configuration file:

1. https://github.com/lsst-ts/ts_wep/blob/develop/policy/cwfs/instData/lsst/instParamPipeConfig.yaml
2. https://github.com/lsst-ts/ts_wep/blob/develop/policy/cwfs/instData/comcam/instParamPipeConfig.yaml
3. https://github.com/lsst-ts/ts_wep/blob/develop/policy/cwfs/instData/lsstfam/instParamPipeConfig.yaml

python/lsst/ts/ofc/sensitivity_matrix.py
```diff
@@ -0,0 +1,152 @@
...
+                )
+                for dof_idx in self.ofc_data.dof_idx
+            ]
+        )
+        # Reshape the sensitivity matrix to have the
+        # same shape as the one stored.
+        rotated_sensitivity_matrix = np.moveaxis(rotated_sensitivity_matrix, 0, -1)
+
+        # Subselect the relevant zernike coefficients
+        # to include in the sen matrix.
```
teweitsai, line 118: sen -> sensitivity

python/lsst/ts/ofc/sensitivity_matrix.py
```diff
@@ -0,0 +1,152 @@
...
+        List of field x angles in degrees.
+    field_y : `list` of `float`
+        List of field y angles in degrees.
+    """
+
+    # Create camera object
+    camera = obs_lsst.LsstCam().getCamera()
+
+    # Get the field angle for each sensor
+    fieldList = []
```
teweitsai, line 146: fieldList = list()

tests/test_sensitivity_matrix.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def setUp(self):
+        self.ofc_data = OFCData("lsst")
+
+        self.sensor_names = ["R44_SW0", "R04_SW0", "R00_SW0", "R40_SW0"]
+
+    def test_sensitivity(self):
+        sensitivity_matrix = SensitivityMatrix(0.0, self.sensor_names)
+
+        print(sensitivity_matrix)
```
teweitsai, line 41: Please use the assertion function instead of `print` to do the test. In addition, you should provide the test of different rotation angle and judge the result makes sense or not.

tribeiro (changes requested):
Please! don't add fits files as normal files! If you want to add fits files to the repo, add them as git lfs.

jmeyers314:
> Remove all the sensitivity fits file and use the yaml instead. The fits file is binary and makes the git diff useless. If you need to hold the data as fits in memory (although I guess it is easier to hold it in numpy.array, you could change the yaml to fits in the runtime.

While I appreciate the desire for quick diffs, and I think yaml is great for short data files where I can pretty quickly grok what the line # actually means, I pretty strongly disagree with yaml being an appropriate format in this case.  

I just ran a quick check on policy/lsstfam/senM_gq_189_19_50.fits, and it takes about 6-7 ms to load as a fits file (or ~18 ms if gzipped).  If I convert it to yaml (of either flow-style) and load that, it's about 5700-6500 ms.  That 1000x difference feels pretty unacceptable to me.   Six seconds is a long time when you're debugging or running unit tests.  And it's a hefty fraction of our 20-40s we get for the closed loop if we ever need to restart ts_ofc.  And even if git diff showed a change at, say, line 14235 out of 23207, I'd pretty much need to open the data as a numpy array to understand what field-angle/zernike/dof is effected anyway.  

So I think this should probably stay as a binary file.  FITS is a stable enough format.

git-lfs though makes sense.

python/lsst/ts/ofc/sensitivity_matrix.py
tribeiro: Please! use snake_case,

```
field_list
```

python/lsst/ts/ofc/test.ipynb
tribeiro, line 1: Can you remove this notebook? Or, if you really want to keep it, move it to a `notebooks` directory in the top level. But I would rather delete it. 

If you want to document some usage, do it in the doc page with the doc format.

python/lsst/ts/ofc/ofc.py
```diff
@@ -151,13 +151,13 @@ def calculate_corrections(self, wfe, field_idx, filter_name, gain, rot):
         else:
             self.ofc_controller.gain = gain
 
-        optical_state = self.state_estimator.dof_state(filter_name, wfe, field_idx)
+        optical_state = self.state_estimator.dof_state(filter_name, wfe, field_idx, rot)
```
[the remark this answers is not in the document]
gmegh: It is consistent. The rotation angle is passed in this function to compute the rotated sensitivity matrix and that is done only when we use the Double_zernikes attribute. Otherwise (when DZ is not enabled), this argument here is not used. When DZ is not enabled we rotate the degree of freedom state. When it is enabled we only rotate the sensitivity matrix. Let me know if you need further clarification.

teweitsai:
If we want to use the `fits` to hold the sensitivity matrix, then, we would need to have a way to document the related details and do the related maintenance. The benefit of `yaml` is that you can put the comment to explain the related details. The sensitivity matrix contains the information of basis and unit. For the future's developer, it would be hard for him/her to guess what is the detail for the encoded value. He/she will only know that is a 3-D matrix and nothing more than that. But actually, even you document this in the `doc/` or `README.md`, it would still be easy to be out of sync if the developer in future decides to change the definition in sensitivity matrix.

In addition, in the future, we might just fine-tune some values in the sensitivity matrix based on the experimental data on summit. Under this condition, the `git diff` can know the change immediately. Therefore, unless there is a good way to solve the maintenance problem in the software side, I think the `yaml` might be a better choice compared with `fits`.

For the performance issue, it might be annoying in the simulation level. However, for the control side, the `ts_MTAOS` will read and hold all the data when start the CSC. Since everything will be hold in memory, the performance would not be an issue in the control side.

python/lsst/ts/ofc/ofc.py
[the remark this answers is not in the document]
teweitsai: In this case, you should explain this argument and related detail in the doc string of `StateEstimator.dof_state()`. In addition, it sounds like this should be an `optional` argument.

jmeyers314:
@teweitsai , Hmm... I think a README inside the relevant policy subdirectories (not the main repo README.md) + git history would actually be fine for maintenance.  

But, if you're worried about things getting desynchronized, then maybe we ought to consider [asdf](https://en.wikipedia.org/wiki/Advanced_Scientific_Data_Format).  It's apparently used by JWST, and more or less consists of a mashup of FITS and yaml.  I checked and it's already available in the science pipelines, though I don't know who in DM is actually using it.  It's also comparable in speed to bare FITS, and we could put whatever commentary/history we want in the yaml portion of the file.  The header is human-readable / viewable with unix tools, e.g., `head myfile.asdf -n20`.  If JWST is using it, then that would be evidence enough for me for stability.  Only potential downside I can see is `git diff` probably doesn't work; at least regular diff claims my test asdf file is binary and won't display the actual differences, even if they're only in the header.  However, I'm skeptical that anybody would want to run diff frequently enough that this should be considered an impediment to using a binary file.  The speed gain, multiplied by the thousands of times this will get loaded over time is definitely worth it in my opinion.

teweitsai:
@jmeyers314 The use of `asdf` sounds a good solution to us for the maintainability and performance. Thanks!

tribeiro (commented):
This is looking pretty good. I have a few inline comments but nothing major. I think you can take this out of draft mode. once you go over my first round of comments I'll do another round and I think we will be ready to merge. 

Also, please, make sure you run black and flake8, I noticed some clear formatting issues.

python/lsst/ts/ofc/scripts/generate_sensitivity_matrix.py
```diff
@@ -0,0 +1,362 @@
+import argparse
+from astropy.io import fits
+import batoid
```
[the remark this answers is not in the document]
jmeyers314: I suspect the intention here is more "document where the sensitivity matrix came from" as opposed to "here's a thing we expect users to run"?  If so, then I don't think we need to have ts_ofc depend on batoid or galsim.  Maybe this file needs to live in a different part of the repo then though?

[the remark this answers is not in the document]
gmegh: Just to add to this discussion, galsim will be needed regardless because we use it to evaluate the double zernike object. I think batoid might not be needed. I defer to Tiago to where it should be placed if not here

policy/sample_points/lsst/img_quality_weight.yaml
```diff
@@ -34,7 +34,3 @@
 "28": 0.2369
 "29": 0.2369
 "30": 0.2369
-"31": 0.25
```
jbkalmbach: What do you think about keeping these in here and instead removing the code in `ts_imsim` that adds them in?

jbkalmbach: Because right now it happens in a simulation step and I think I'd rather have OFC handle it for the real data.

jbkalmbach: Also maybe we set the values to be 1.0 since they will be normalized anyway.

conda/meta.yaml
```diff
@@ -32,3 +35,8 @@ requirements:
     - setuptools
     - setuptools_scm
     - scipy >=1.6
+    - batoid >= 0.5
+    - galsim >= 2.5
+    - ruamel.yaml
+    - obs-lsst
```
tribeiro, line 41: I don't think this package exists.

python/lsst/ts/ofc/ofc.py
```diff
@@ -76,13 +75,13 @@ class OFC:
     PSSN: Normalized point source sensitivity.
     """
 
-    def __init__(self, ofc_data, log=None):
+    def __init__(self, ofc_data: OFCData, log: logging.Logger = None) -> None:
```
tribeiro, line 78: If `log` can be of either `logging.Logger` or `None`, the type should be `logging.Logger | None`, e.g.:

```
 def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None:
```

python/lsst/ts/ofc/ofc.py
```diff
@@ -76,13 +75,13 @@ class OFC:
...
         if log is None:
             self.log = logging.getLogger(type(self).__name__)
         else:
             self.log = log.getChild(type(self).__name__)
 
-        self.pssn_data = dict(sensor_id=None, pssn=None)
+        self.fwhm_data = None
```
tribeiro, line 84: Aren't you doing this in  #54 ? I think this topic is a bit contentious no?

python/lsst/ts/ofc/ofc.py
```diff
@@ -91,16 +90,25 @@ def __init__(self, ofc_data, log=None):
...
 
-        self.cam_rot = CamRot()
-
         self.dof_order = ("m2HexPos", "camHexPos", "M1M3Bend", "M2Bend")
 
-    def calculate_corrections(self, wfe, field_idx, filter_name, gain, rot):
+    def calculate_corrections(
+        self,
+        wfe: np.ndarray,
+        sensor_names: list,
```
tribeiro, line 107: Is this a list of strings? If so type should be `list[str]`

python/lsst/ts/ofc/ofc.py
gmegh: Removed these changes from this PR

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -350,122 +314,207 @@ def _configure_instrument(self, value):
...
-                f"Could not read image quality file from instrument config directory: {iqw_path!s}. "
+                "Could not read weights file from instrument"
+                f"config directory: {image_quality_weights_path!s}. "
                 "Check your instrument configuration directory integrity."
             )
 
+        # If the camera type is lsst read and set up
+        # gaussian quadrature weights
+        # ------------------------------------------
+        if camera_type == "lsst":
```
tribeiro, line 389: I am not sure I understand the need for this distinction. can you clarify?

tribeiro (commented):
Hear some additional inline comments. However I have a major concern about this PR I would like to voice here.

I am very skeptical we will be able to converge on this PR in a timely fashion. 

I still have a hard time understanding what is related to the double zernikes and what is not. 

In the PR description you say:

> This PR addresses the rotation issues in ts_ofc along with various other issues that needed to be fixed.

That "_along with various other issues that needed to be fixed_" is killing us. It is very hard to focus on what is important if we have to jiggle between changes that are unrelated. 

Here is what I propose we do:

1. Get a list of all the "various other issues that needed to be fixed" and make individual tickets for them. If they are related you could have 1 ticket, but since you say "various" I think it means several tickets.
2. Close this PR without merging it. It is way past the discussion point where I think we will be able to re-focus it.
3. If you can make a PR with just the double zernikes, without fixing the various issues you pointed above, then, refactor the branch to contain only the double zernikes and open a new PR. If you need the fixes before you do the changes, then get the PRs in with the fixes first, then make a PR with the double zernikes only.

gmegh:
Closing this PR as suggested by Tiago. Will be opening another one with the changes in it and requesting only @teweitsai comments

PR #52, opened 2023-07-21
Tickets/DM-40089

suberlak:
Fix setup files to follow https://github.com/lsst-ts/ts_m2com/blob/develop/setup.cfg in accord with https://tssw-developer.lsst.io/procedures/pre_commit.html#pre-commit


couger01 (approved):
LGTM

PR #54, opened 2023-11-03
Removed PSSN usage for FWHM

gmegh:
This PR removes calculations of PSSN that weren't actually used. Up until now, we were passing the FWHM values to calculate the PSSN values, and we would then calculate the FWHM values back from the PSSN to obtain the gain. This PR removes the PSSN completely from this repository as it is not needed. Ts_imsim still stores the values of PSSN but it has its own function to compute it. 

Additionally, I removed the usage of image quality weight, because it was only used to do an average, both in the case of `lsstfam`, in which all the points have the same weight, and the case of `lsst` in which case we only use the four corners to set the gain and we give them an equal gain.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,13 @@
 Version History
 ##################
 
+.. _lsst.ts.ofc-2.1.2:
+
+v2.1.2
+======
+
+* Substituted pssn to fwhm for gain calculations.
+
 .. _lsst.ts.ofc-2.1.1:
 
 v2.1.1
--- a/python/lsst/ts/ofc/ofc.py
+++ b/python/lsst/ts/ofc/ofc.py
@@ -49,13 +49,13 @@ class OFC:
     cam_rot : `CamRot`
         Camera rotator class.
     default_gain : `float`
-        Default gain, used when setting gain in `set_pssn_gain()` when fwhm is
-        above `fwhm_threshold`.
+        Default gain, used when setting gain in
+        `set_gain_from_fwhm()` when fwhm is above `fwhm_threshold`.
     dof_order : `tuple`
         Order of the degrees of freedom.
     fwhm_threshold : `float`
         Full width half maximum threshold when estimating gain with
-        `set_pssn_gain()`.
+        `set_gain_from_fwhm()`.
     log : `logging.Logger`
         Logger class used for logging operations.
     lv_dof : `np.ndarray`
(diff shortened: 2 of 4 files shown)


teweitsai:
The current values of image quality weight are just a place holder at this moment. Actually, the `lsstfam` should not have the same values for all the sensors (Bo promised before to evaluate them but he was too busy). Therefore, I believe we should keep them for the tuning at the commissioning stage.

tribeiro:
I agree with TeWei, I am not sure I agree with you that we won't use this feature. I would like to better understand the reasons for removing it before we actually do it.

gmegh:
> the tuning of gain value should base on the PSSN instead of FWHM, which provides more physical information.

If you look at the current code, the PSSN is calculated from the FWHM and then we compute back the FWHM from the PSSN. That does not eeally makes sense so I believe we should just be using the FWHM. The relation between both is analytical, by the way.

i understand the weights point, and I can add back that feature. 

That said, please re-look at the old pssn / fwhm functions, they were messy and we weren’t actually using the pssn for any purpose.

teweitsai:
I think the original design is this:

1. The offline process in DM's side will publish the FWHM regularly as event or data by butler (middleware).
2. The MTAOS CSC will subscribe these FWHM data and calculate the PSSN based on an internal physical/atmosphere model (`OFCController.fwhm_to_pssn()`).
3. The OFC calculates the expected FWHM with this internal model (`OFCController.effective_fwhm_g4()`). This means this FWHM contains the atmosphere information.
4. Then, this new FWHM value is used to decide the `gain` value (`OFC.set_pssn_gain`).

Although it sounds like the FWHM is the only needed input, the PSSN can reflect the physics behind better.
The current transformation between FWHM and PSSN is based on a very simple atmosphere model that should be able to improve more once we have the experimental data to construct a semi-empirical model when needed.

gmegh:
If you could please show mathematically what you mean I would appreciate
it. What you are saying seems to me the following:

We have a model of the atmosphere with which we will compute the PSSN from
DM’s FWHM. Then with exactly the same model, we will compute the FWHM back
from the PSSN. Therefore the original FWHM will match this computed FWHM,
because we are just using the same model, and that does not make sense to
me.

teweitsai:
I think I know the confusion now. Actually, the `OFCController.fwhm_to_pssn()` is just a place holder at this moment, which needs to develop a semi-empirical model. The following is the Bo's previous implementation:

1. https://github.com/bxin/IM/blob/master/source/aosMetric.py#L333-L446
2. https://github.com/bxin/IM/blob/master/source/aosMetric.py#L222-L223


You can see Bo is based on the atmosphere model and OPD in PhoSim to calculate the PSSN, but we could only get the FWHM as an index of image quality in the initial beginning. Therefore, the easiest way to be a placeholder is to inverse the calculation at `OFCController.effective_fwhm_g4()`, which you can find the related formula in Bo's document relates to PSSN calculation.

PR #57, opened 2024-03-25
Extending up to zk 28

gmegh:


python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+
+import argparse
+import pathlib
+from copy import copy
+
+import batoid
+import galsim
+import numpy as np
+import yaml
+from lsst.ts.imsim import OpdMetrology
```
teweitsai, line 30: If you only needs the `calc_pssn()` from `ts_imsim`, please put it into this `ts_ofc` and let the `ts_imsim` to call it. In the summit deployment, I think we might not deploy the `ts_imsim` (only simulation purpose) to minimize the space.

In addition, the software dependency (or flow) was designed to be `ts_wep` --> `ts_ofc` --> `ts_imsim`. But if you let `ts_ofc` to call the function of `ts_imsim`, you will get the circular dependency, which usually we tried to avoid as much as possible. The circular dependency would give a lot of trouble in the final.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+        header = (
+            "---\n"
+            "\n# Default alpha coefficient for the normalized point-source sensitivity\n"
+            "# Normalized point-source sensitivity (PSSN) information\n"
+            "# PSSN ~ 1 - alpha * delta^2\n"
+            "# Eq (7.1) in Normalized Point Source Sensitivity for LSST\n"
+            "# There are 25 terms of alpha value here for z4-z28 to use.\n"
+            "# The procedure to obtain them is detailed in scripts/generate_alpha_values.py\n"
+            "# For more details, please check the page 19-21 in:\n"
+            "# https://docushare.lsst.org/docushare/dsweb/Get/Document-18041/150513.pptx.pdf\n\n"
```
teweitsai, line 64: Yep, please help to put the unit of alpha here.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+    """Calculate alpha values for the normalized point-source sensitivity.
+
+    Parameters
+    ----------
+    config : dict
+        Configuration dictionary.
+    instrument : str
+        Instrument name.
+    band : str
+        Filter band.
```
teweitsai, line 86: Please use the enum instead:
https://github.com/lsst-ts/ts_wep/blob/develop/python/lsst/ts/wep/utils/enumUtils.py#L56-L63

But I saw this:
https://github.com/lsst-ts/ts_ofc/pull/57/files#diff-f30057d41d35ea471822e3136c9a76ddfe777738a514f285607aabc3b1d83b02R243-R247

Therefore, it should be fine.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+    znmax: int = 28,
+) -> np.ndarray:
+    """Calculate alpha values for the normalized point-source sensitivity.
+
+    Parameters
+    ----------
+    config : dict
+        Configuration dictionary.
+    instrument : str
+        Instrument name.
```
teweitsai, line 84: Please list the available instruments here.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+    znmax : int
+        Maximum Zernike index.
+
+    Returns
+    -------
+    np.ndarray
+        Alpha values for the normalized point-source sensitivity.
+    """
+
+    # Set wavlength and number of zernikes used
```
teweitsai, line 98: wavlength -> wavelength

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+    num_zk = znmax - znmin + 1
```
teweitsai, line 99: If someone set `znmax` to be smaller than `znmin`, the code will die. You may need to explain this in the doc string or add a check and throw out an error here.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+    wavelength = config["wavelengths"][band]
+
+    # Set the increments of zernikes used to do the fit
+    increments = np.array([2, 4, 6, 8, 10]) * wavelength * 1e-2
```
teweitsai, line 103: Please help to explain the magic numbers 2, 4, 6, 8, 10 here.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+
+    # Get the field angles of the gaussian quadrature points
+    field_angles = [
+        ofc_data.gq_points[sensor] for sensor in range(len(ofc_data.gq_weights))
+    ]
+    field_x, field_y = zip(*field_angles)
+
+    # Load the gaussian quadrature weights
+    imqw = [ofc_data.gq_weights[sensor] for sensor in range(len(ofc_data.gq_weights))]
+    n_imqw = imqw / np.sum(imqw)
```
teweitsai, line 122: This line assume the weights (`imqw`) are all positive values: 

```
 n_imqw = imqw / np.sum(imqw)
```

It is not very uncommon to have the negative weight in physics:
https://www.physicsforums.com/threads/negative-weights-in-weighted-averages-explained.429827/

You could consider to throw out an error if there is any negative weight or have a more robust way to deal with it in physics based on your context.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+
+            for id_angle in range(len(field_angles)):
+                # Convert field angles to radians
+                field_x = np.deg2rad(field_angles[id_angle][0])
+                field_y = np.deg2rad(field_angles[id_angle][1])
+
+                # Calculate the wavefront without perturbation
+                # to retrieve the opd mask
+                opd_w = batoid.wavefront(
+                    fiducial, field_x, field_y, wavelength=wavelength, nx=255
```
teweitsai, line 145: Is 255 means the pixel (256 x 256)? If yes, please explain it in the comment.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+                )
+                opd_size = opd_w.array.shape[0]
+                opd_grid_1d = np.linspace(-1, 1, opd_size)
+                opd_x, opd_y = np.meshgrid(opd_grid_1d, opd_grid_1d)
+                non_masked = ~opd_w.array.mask
+
+                # Set zernikes for the new opd
+                new_opd = copy(opd_w.array)
```
teweitsai, line 153: what is the size/shape of `opd_w.array`? I am not sure you need to do `copy()` or `deepcopy()` here? If this `array` is a `numpy.array`, there should be a `copy()` function already.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+
+    Parameters
+    ----------
+    args : argparse.Namespace
+        Arguments passed from command line.
+
+    Raises
+    ------
+    RuntimeError
+        IF the configuration file for the specified instrument is not found.
```
teweitsai, line 204: IF --> If

python/lsst/ts/ofc/scripts/generate_alpha_values.py
```diff
@@ -0,0 +1,271 @@
...
+        default=28,
+        type=int,
+        help="Maximum zk index",
+    )
+
+    return parser.parse_args()
+
+
+def generate_intrinsic_zernikes() -> None:
+    main(parse_arguments())
```
teweitsai, line 267: Please add the doc string.

teweitsai (changes requested):
Looks good. Please consider:

1. Circular dependency of packages: `ts_ofc` vs `ts_imsim`.
2. Inconsistent calculation between the document and code (`alpha`, one has the square and the other one does not ). Please explain the disappearance of `2*pi` as well.
3. Incomplete documents, especially for the magic numbers and new attribute.
4. For the new attribute (`self.alpha` if I was not wrong), please put to `__init__()` to let other developers know there is this attribute in this class.
5. Unit of `alpha` in `yaml` file.

Thanks!

policy/alpha_values.yaml
```diff
@@ -0,0 +1,36 @@
+---
+
+# Default alpha coefficient for the normalized point-source sensitivity
+# Normalized point-source sensitivity (PSSN) information
+# PSSN ~ 1 - alpha * delta^2
+# Eq (7.1) in Normalized Point Source Sensitivity for LSST
+# There are 25 terms of alpha value here for z4-z28 to use.
+# The procedure to obtain them is detailed in scripts/generate_alpha_values.py
+# For more details, please check the page 19-21 in:
+# https://docushare.lsst.org/docushare/dsweb/Get/Document-18041/150513.pptx.pdf
```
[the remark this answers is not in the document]
gmegh: Added the units. While the extension is weird, it's the URL to the presentation that has the information on how to get the alpha values

python/lsst/ts/ofc/ofc_controller.py
```diff
@@ -412,10 +412,7 @@ def uk(
         #         = x.T * (A.T * C.T * C * A) * x = x.T * Q * x
         # CCmat is C.T *C above.
 
-        cc_mat = (
-            2.0 * np.pi / self.ofc_data.eff_wavelength[filter_name]
-        ) ** 2.0 * self.ofc_data.alpha[self.ofc_data.zn3_idx]
-        cc_mat = np.diag(cc_mat)
+        cc_mat = np.diag(self.ofc_data.alpha[self.ofc_data.zn_idx])
```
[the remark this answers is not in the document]
gmegh: Updated the math to show diag(\alpha), the new alpha values are already in um^-2, which is a more natural unit than the rad^2 that were there before.

python/lsst/ts/ofc/scripts/generate_alpha_values.py
gmegh: I see your point. Moving calc_pssn into ts_ofc would mean moving this "https://github.com/lsst-ts/ts_imsim/blob/develop/python/lsst/ts/imsim/utils/metro_tool.py" all to ts_ofc. I can do that, however I would like to point out a counter-argument. This function will not be used often, that is why is placed in the scripts directory. For the generate_sensitivity_matrix script, for instance, we decided to keep it in ts_ofc even though it imports "batoid_rubin" which is not available in ts_ofc. It is just to have them in the same repository so that if someone needs to regenerate these values, they can. In that sense, given the sporadic use of this script one could argue that it's not really circular. 

But if you have a strong opinion on this, I am happy to move all the metro_tool.py over here, I am just worried it will be a lot of code for a function that we only use once or twice and not in the closed-loop.

doc/developer-guide/control-algorithm/ca-in-ofc.rst
```diff
@@ -96,10 +96,10 @@ This means :math:`\textbf{Q} = \textbf{A}^{T}\textbf{C}^{T}\textbf{C}\textbf{A}`
 The matrix :math:`\textbf{C}^{T}\textbf{C}` is chosen to be:
 
 .. math::
-    \textbf{C}^{T}\text{C} \equiv \left( \frac{2\pi}{\lambda(\ell)} \right)^{2} \text{diag}(\vec\alpha)^{2},
+    \textbf{C}^{T}\text{C} \equiv \text{diag}(\vec\alpha)^{2},
 
 where :math:`\vec{\alpha}` is the alpha value of normalized point source sensitivity (PSSN) in the basis :math:`\mathbb{Z}`.
-:math:`\lambda(\ell)` is the effective wavelength under different filter band :math:`\ell`.
+The values are computed by fitting :math:`1 - PSSN = \alpha_k \delta_k^2`, where :math:`\delta_k` is the Zernike coefficient perturbation in microns. The units of :math:`\alpha_k` are in :math:`\mu m^{-2}`.
```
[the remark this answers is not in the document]
gmegh: 1. Fixed
2. The term 2pi/lambda was a vestige of the old alpha values which were in rad^-2, as can be seen [here](https://docushare.lsst.org/docushare/dsweb/Get/Document-18041/150513.pptx.pdf). But I think it's more natural to have them in um^-2, which is the unit of zernikes squared. Having alpha in um^-2 removes the need to have the 2pi/lambda, which was basically a unit change

doc/versionHistory.rst
```diff
@@ -13,6 +13,7 @@ v3.1.0
 
 * Update sensitivity matrix and intrinsic zernikes up to zk 28.
 * Add zn_selected to select the zernikes to be used.
+* Updated alpha values for image quality matrix
```
teweitsai, line 16: Period at the end.

teweitsai (approved):
Looks good to me now. Thanks for the update!

python/lsst/ts/ofc/scripts/generate_alpha_values.py
teweitsai: The cycle dependency would be a disaster for the software maintenance.
Please discuss with UW team to decide to put this `metro_tool.py` to the `ts_wep` or `ts_ofc` for the downstream packages to use (this depends on the `ts_wep` will need it or not).
This can be a new DM ticket.
As a reference, you can see: https://en.wikipedia.org/wiki/Circular_dependency

Although I could understand the point here that this function will be rarely used, I still believe to avoid the cycle dependency is a good software habit, which can decrease the subtle bugs in the future.

teweitsai:
BTW, at the next time of PR, please consider to reorganize the history before issuing the PR:
https://medium.com/@catalinaturlea/clean-git-history-a-step-by-step-guide-eefc0ad8696d

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -481,6 +517,7 @@ def _configure_instrument(self, instrument):
         self.log.debug(f"Done configuring {instrument}")
 
         # Now all data was read successfully, time to set it up.
+        self.alpha = alpha
```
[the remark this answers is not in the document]
gmegh: So, in this case I actually think it should not be declared in init, just like all these other variables are not declared: 
```
self.alpha = alpha
        self.config = config
        self.dof_state0 = dof_state0
        self.sample_points = sample_points
        self.gq_points = gq_points if instrument == "lsst" else None
        self.image_quality_weights = image_quality_weights
        self.gq_weights = gq_weights if instrument == "lsst" else None
        self.y2_correction = y2_correction
        self.gq_y2_correction = gq_y2_correction if instrument == "lsst" else None
        self.intrinsic_zk = intrinsic_zk
```

Alpha is a variable that should only be set if we configure the instrument. Let me know your thoughts!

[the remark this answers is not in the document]
teweitsai: Other variables are declared in the `BaseOFCData` (a `dataclass`) actually (the way here is like the `__init__()` to declare the attributes):

1. https://github.com/lsst-ts/ts_ofc/blob/37adfaf37b6be97fc505bc9e6409beaa7388cb09/python/lsst/ts/ofc/ofc_data/base_ofc_data.py#L101-L137
2. https://docs.python.org/3/library/dataclasses.html

The `alpha` was there originally but was removed in this PR:

https://github.com/lsst-ts/ts_ofc/pull/57/files#diff-a12fd56ca593e8d61e1552dd2c58e9d30ac8360e2131fe426c945bca1ce19da9R133-R135

Therefore, you can put the above line back and do something like the following to declare it:

```
alpha: np.ndarray = np.array([])
```

Or you can modify the code of `BaseOFCData.default_alpha()` (removed in this PR) to read the `yaml` file.

The rule is that the attributes of class should be declared at the expected place, and developers know where to find them in the `dataclass` or `class.__init__()`.

gmegh:
After fixing the penalty term to balance the increased alpha values, convergence is reached as it was before. See plot attached below. 

Screenshot 2024-04-05 at 13 44 19

teweitsai:
Thanks for this effort! The convergence looks good. But I think you could open another ticket to tune the parameters because the iterations 3-8 (z28) have the oscillations which implies something might not be right in the control part. Thanks!

PR #60, opened 2024-05-21
Move class diagrams to mermaid from plantUML.

jbkalmbach:


doc/uml/ofc_class.mmd
```diff
@@ -13,4 +14,4 @@ OFC *-- SensitivityMatrix
 OFC *-- StateEstimator
 OFC *-- OFCController
 OFC ..> Correction
-@enduml
+OFCData --|> BaseOFCData
```
teweitsai, line 17: I think usually we will do: `BaseOFCData <|-- OFCData` as a convention.

teweitsai:
You can remove the `PLANTUML_URL` and related command in `Jenkinsfile` as well.

PR #62, opened 2024-07-09
Add PID option to OFC

gmegh:
This PR includes:
- Creates a directory of controller classes to be used by OFC. 
- Removes OFCController and substitutes it by OICController and PIDController
- These new controllers are configured through a yaml configuration file to tweak gains

### DIFF
--- a/python/lsst/ts/ofc/controllers/base_controller.py
+++ b/python/lsst/ts/ofc/controllers/base_controller.py
@@ -1,0 +1,383 @@
+# This file is part of ts_ofc.
+#
+# Developed for Vera Rubin Observatory.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import logging
+
+import numpy as np
+
+from .. import OFCData
+
+__all__ = ["BaseController"]
+
+
+# TODO: DM-45103 Add saturator once bending mode limits are determined.
+class BaseController:
+    """Base Controller.
+
+    This class is the base class for all the controllers.
+    This class is mainly used to determine the offset (`uk`) of degree of
+    freedom (DOF) at time `k+1` based on the wavefront error (`yk`) at time
+    `k`.
+
+    Parameters
+    ----------
+    ofc_data : `OFCData`
+        OFC Data.
+    log : `logging.Logger` or `None`
+        Optional logging class to be used for logging operations. If `None`,
+        creates a new logger.
+
+    Attributes
+    ----------
+    dof_state : `np.array`
+        State of telescope in the basis of degrees of freedom.
+    dof_state0 : `np.array`
+        Initial state of telescope in the basis of degrees of freedom.
+    integral : `np.array`
+        Integral of the error.
+    kp : `float`
+        Proportional gain.
+    ki : `float`
+        Integral gain.
+    kd : `float`
+        Derivative gain.
+    log : `logging.Logger`
+        Logger class used for logging operations.
+    ofc_data : `OFCData`
+        OFC data.
+    previous_error : `np.array`
+        Previous error.
+    pssn_data : `dict`
+        PSSN data.
+    setpoint : `np.array`
+        Setpoint for the PID controller.
+    """
+
+    # Eta in FWHM calculation.
+    ETA = 1.086
+
+    # FWHM in atmosphere.
+    FWHM_ATM = 0.6
+
+    def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None:
+        if log is None:
+            self.log = logging.getLogger(type(self).__name__)
+        else:
+            self.log = log.getChild(type(self).__name__)
+
+        # Set OFC data
+        self.ofc_data = ofc_data
+
+        # Set gains and setpoint
+        self._kp = self.ofc_data.controller["kp"]
+        self._ki = self.ofc_data.controller["ki"]
+        self._kd = self.ofc_data.controller["kd"]
+        self._derivative_filter_coeff = self.ofc_data.controller[
+            "derivative_filter_coeff"
+        ]
+        self.setpoint = np.array(self.ofc_data.controller["setpoint"])
+
+        # Set initial state
+        self.dof_state0 = np.zeros(len(self.ofc_data.dof_idx))
+
+        for comp in self.ofc_data.comp_dof_idx:
+            start_idx = self.ofc_data.comp_dof_idx[comp]["startIdx"]
+            end_idx = (
+                self.ofc_data.comp_dof_idx[comp]["startIdx"]
+                + self.ofc_data.comp_dof_idx[comp]["idxLength"]
+            )
+
+            state0name = self.ofc_data.comp_dof_idx[comp]["state0name"]
+
+            if "Hexapod" in state0name:
+                self.dof_state0[start_idx:end_idx] = np.array(
+                    [
+                        self.ofc_data.dof_state0[state0name][axis_name]
+                        for axis_name in ["dZ", "dX", "dY", "rX", "rY"]
+                    ]
+                )
+            else:
+                self.dof_state0[start_idx:end_idx] = np.array(
+                    [
+                        self.ofc_data.dof_state0[state0name][f"mode{mode_idx+1}"]
+                        for mode_idx in range(
+                            self.ofc_data.comp_dof_idx[comp]["idxLength"]
+                        )
+                    ]
+                )
+
+        self.dof_state = self.dof_state0.copy()
+
+        # Initialize previous error and integral
+        self.previous_error = self.setpoint - self.dof_state0
+        self.integral = self.previous_error
+        self.filtered_derivative = np.zeros(len(self.ofc_data.dof_idx))
+
+        # Initialize PSSN data
+        self.pssn_data = dict(sensor_names=None, pssn=None)
+
+    @property
+    def kp(self) -> float:
+        """Proportional gain."""
+        return self._kp
+
+    @kp.setter
+    def kp(self, value: float) -> None:
+        self._kp = value
+
+    @property
+    def ki(self) -> float:
+        """Integral gain.
+
+        Returns
+        -------
+        `float`
+            Integral gain.
+        """
+        return self._ki
+
+    @ki.setter
+    def ki(self, value: float) -> None:
+        self._ki = value
+
+    @property
+    def kd(self) -> float:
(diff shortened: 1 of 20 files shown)


gmegh:
I will attach here a screenshot of the closed loop results from ts_imsim with OIC and PID in a bit!

teweitsai (changes requested):
Please help:

1. Update the UML diagram: [ofc_class.mmd](https://github.com/lsst-ts/ts_ofc/blob/develop/doc/uml/ofc_class.mmd)
2. Add the `mypy` support. I noticed some codes have the `mypy` check and some do not.
3. Provide the function to clear the PID history. I do not find this in `base_controller.py`.
4. Since the PID is used, usually, in the control system, people will add the filter and saturator to avoid the "pd" term to explode. If you think these two items are not needed, please help to provide the arguments to support this.

Thanks!

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -0,0 +1,305 @@
...
+        self.pssn_data = dict(sensor_names=None, pssn=None)
+
+    @property
+    def kp(self):
+        return self._kp
+
+    @kp.setter
+    def kp(self, value):
+        if 0.0 <= value <= 1.0:
+            self._kp = value
```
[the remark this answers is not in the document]
gmegh: I have removed this constraint and the corresponding unit test function that tested it.

python/lsst/ts/ofc/controllers/pid_controller.py
```diff
@@ -0,0 +1,98 @@
...
+    Parameters
+    ----------
+    kp : `float`
+        Proportional gain.
+    ki : `float`
+        Integral gain.
+    kd : `float`
+        Derivative gain.
+    setpoint : `float`, optional
+        Setpoint for the controller.
```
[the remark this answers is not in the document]
gmegh: That's because those attributes are set in the parent class. Should I then remove them from this docstring, even though they are also attributes of the PIDController?

python/lsst/ts/ofc/controllers/oic_controller.py
```diff
@@ -292,7 +120,7 @@ def calc_uk_x0(self, mat_f, qx, **kwargs):
         uk : `numpy.ndarray`
             Calculated uk in the basis of degree of freedom (DOF).
         """
-        return -mat_f.dot(qx)
+        return mat_f.dot(qx)
```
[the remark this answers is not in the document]
gmegh: That's because with the new implementation we use that value and then do `kp * error`, where error is `error = setpoint - state` and in this case `state = mat_f.dot(qx)`

python/lsst/ts/ofc/controllers/base_controller.py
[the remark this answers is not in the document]
teweitsai: Yep. The following link should explain why we could remove the restriction:

[File:PID Compensation Animated.gif](https://en.wikipedia.org/wiki/File:PID_Compensation_Animated.gif)

And since we might have the overshoot in the tuning of PID parameters, it is common to add the saturator. For example, you could add the restriction for the final output `uk` should be less than the `error` or 2 * `error` or others (you can put the parameter to avoid to hard-code anything).

Thanks!

python/lsst/ts/ofc/controllers/oic_controller.py
```diff
@@ -498,24 +290,83 @@ def uk(
...
+        # Add integral term if ki is not None
+        if self.ki is not None:
+            self.integral += current_error
+            control_effort += self.ki * self.integral
+
+        # Add derivative term if kd is not None
+        if self.kd is not None:
+            derivative = current_error - self.previous_error
+            self.previous_error = current_error
+            control_effort += self.kd * derivative
```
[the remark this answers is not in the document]
teweitsai: I checked again your implementation and realized you are using the idea of 2.5.1 in the following page:

https://www.betzler.physik.uni-osnabrueck.de/Manuskripte/Elektronik-Praktikum/p3/doc2558.pdf

That means your `ki` and `kd` here contains the information of `delta T` already. In this case, you may need to describe this clearly in the configuration file when people turn the parameters. I do not know people will turn the parameters based on the Bode diagram or not in the measurement. That means, you need to document all parameters physical meaning clearly.

Thanks!

python/lsst/ts/ofc/controllers/pid_controller.py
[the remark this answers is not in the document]
teweitsai: If the `Child.__init__()` and `Parent.__init__()` have the same signatures and class attributes, you can remove the doc string of `__init__()` from the Child. Otherwise, you may need to copy what you have from the Parent.

As an example, you can do the following experiment:

```python
class Parent()
    """Doc string of parent"""
    def __init__():
        self.a = "foo"

class ChildA(Parent)
    """Doc string of parent"""
    def __init__():
        self.b = "bar"

class ChildB(Parent)
    def function1:
        pass

child_a = ChildA()
child_b = ChildB()

# Do the help for each child to check the doc string and attribute
help(child_a)
help(child_b)
```

And sometimes, based on the condition, we will use the `super()` function in child as well, which depends on the use case.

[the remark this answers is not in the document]
teweitsai: Therefore, in your case, for the `PIDController`, you can remove the function of `__init__()` because it does not add any new logic from the parent. You can try this to verify it.

In the `OICController`, you would need to rewrite all the details of doc string because you overwrote the function (`__init()__`) from the parent with the new attributes.

gmegh:
Since adding `mypy` will create a lot of changes that are unrelated to this PR, I have decided to create a different PR that adds `mypy` support. Hope that's okay!

gmegh:
@teweitsai  Attached are convergence plots
Screenshot 2024-07-16 at 10 58 05
Screenshot 2024-07-16 at 10 58 11
Screenshot 2024-07-16 at 10 58 24
Screenshot 2024-07-16 at 10 57 59

gmegh:
As a note, when noise is present if we don't eliminate the degenerate combinations of degrees of freedom, the PID will underperform the OIC.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -173,44 +199,64 @@ def aggregated_state(self):
         """
         return self.dof_state[self.ofc_data.dof_idx]
 
+    def reset_history(self) -> None:
+        """Reset the history of the controller."""
+        self.dof_state0 = self.dof_state
```
teweitsai, line 224: You might need to do the following instead as the `self.dof_state` is an object:

```
self.dof_state0 = self.dof_state.copy()
```

Please make sure you have a unit test to make sure the `id(dof_state0)` and `id(dof_state)` are different.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -173,44 +199,64 @@ def aggregated_state(self):
...
+        self.integral = (
+            self.setpoint[self.ofc_data.dof_idx] - self.dof_state[self.ofc_data.dof_idx]
+        )
+        self.previous_error = (
+            self.setpoint[self.ofc_data.dof_idx] - self.dof_state[self.ofc_data.dof_idx]
+        )
```
teweitsai, line 230: I am not sure this part of logic. Do you have any (unit) test to verify this? Or for example, run a simulation first. Then, call this function to reset the history, followed by a new simulation to make sure everything in PID will start over. This would be important because we would turn the PID parameters in `ts_mtaos`. We need to make sure before the new turning, we have cleared the history.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -173,44 +199,64 @@ def aggregated_state(self):
...
+        """Estimate uk in the basis of degree of freedom (DOF).
 
         Parameters
         ----------
         filter_name : `string`
             Name of the filter.
         dof_state : `numpy.ndarray`
             Optical state in the basis of DOF.
-        sensor_names : `list` of `string`
+        sensor_names : `list [string]`
```
teweitsai, line 247:
```
sensor_names : `list` [`string`]
```

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -173,44 +199,64 @@ def aggregated_state(self):
...
             List of sensor names.
 
+        Raises
+        ------
+        NotImplementedError
+            Child class should implemented this.
+        """
+        raise NotImplementedError("Child class should implemented this.")
```
teweitsai, line 255: implemented -> implement

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -173,44 +199,64 @@ def aggregated_state(self):
...
+
+    def pid_step(self, state: np.ndarray) -> np.ndarray:
+        """
```
teweitsai, line 258: Please add the description of this function and consider to rename the function name to be `calculate_pid_step()` or `calc_pid_step()`.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -221,7 +267,7 @@ def effective_fwhm_g4(self, pssn, sensor_names):
         ----------
         pssn : `numpy.ndarray` or `list`
             Normalized point source sensitivity (PSSN).
-        sensor_names : `list` of `string`
+        sensor_names : `list [string]`
```
teweitsai, line 302: This comment applies to others. Please check all other codes.

```
sensor_names : `list` [`string`]
```

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -28,6 +28,7 @@
 __all__ = ["BaseController"]
 
 
+# TODO: Add saturator once bending mode limits are determined.
```
teweitsai, line 31: Please put the ticket number DM-xxxxx here.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -250,7 +270,17 @@ def pid_step(self, state: np.ndarray) -> np.ndarray:
         self.integral += error
         derivative = error - self.previous_error
 
-        uk = self.kp * error + self.ki * self.integral + self.kd * derivative
+        # Apply low-pass filter to the derivative term
+        self.filtered_derivative = (
+            self.derivative_filter_coeff * derivative
+            + (1 - self.derivative_filter_coeff) * self.filtered_derivative
+        )
```
teweitsai, line 279: The current implementation is a very simple one. In the future, we might need to abstract this part to a new class to let the users can experiment different filters. In the field of control, sometimes, people would say the control itself is the tuning of filters. Actually, in the frequency domain, you could see the PI controller is a low-pass filter and the PD controller is a high-pass filters as well.

python/lsst/ts/ofc/controllers/pid_controller.py
```diff
@@ -0,0 +1,63 @@
...
+
+__all__ = ["PIDController"]
+
+
+class PIDController(BaseController):
+    """PID controller."""
+
+    def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None:
+        # Initialize base class
+        super().__init__(ofc_data, log)
```
teweitsai, line 37: You can remove this `PIDController.__init__()`.

teweitsai (approved):

teweitsai:
I have a little concern for the implementation to reset the PID history. Please do the simulation test to make sure the history does reset as the expectation. Thanks!

gmegh:
I added the requested unit test. Please comment here if you think the PR is ready and I will squash all the commits. Thanks for all your useful comments!

teweitsai:
I think this PR is ready to merge. Thanks!

PR #63, opened 2024-07-23
Adding mypy support to ofc

gmegh:


python/lsst/ts/ofc/bend_mode_to_force.py
```diff
@@ -94,7 +95,7 @@ def __init__(self, component, ofc_data):
             :, usecols
         ]
 
-    def force(self, dof):
+    def force(self, dof: np.ndarray[float]) -> np.ndarray:
```
teweitsai, line 98: For the return values and others, if you use the `list`, `dict`, `np.ndarray`, or `tuple`, please try to be more specific to say `np.ndarray[float]`, `np.ndarray[int]`, etc. But if this would be too complex to affect the readability, you could ignore it. This comment applies to others. Thanks!

python/lsst/ts/ofc/controllers/oic_controller.py
```diff
@@ -222,7 +224,21 @@ def uk(
...
+        # Otherwise, for full array mode instruments,
+        # we will use the sensor positions to retrieve the y2 correction.
         if self.ofc_data.name == "lsst":
+            if (
+                self.ofc_data.gq_points is None
+                or self.ofc_data.gq_weights is None
+                or self.ofc_data.gq_y2_correction is None
+            ):
+                raise RuntimeError(
+                    "gq_points and gq_weights must be provided for LSST instrument."
```
teweitsai, line 239: Please add the RuntimeError to doc string as well.

python/lsst/ts/ofc/controllers/oic_controller.py
```diff
@@ -231,13 +247,24 @@ def uk(
...
             ]
+            y2c = np.array(
+                [self.ofc_data.gq_y2_correction[idx] for idx in range(len(imqw))]
+            )
         else:
+            if sensor_names is None:
+                raise RuntimeError(
+                    "sensor_names must be provided for full array mode instruments."
+                )
+
```
teweitsai, line 258: Please add the RuntimeError to doc string as well.

python/lsst/ts/ofc/controllers/oic_controller.py
```diff
@@ -349,7 +362,7 @@ def set_pssn_gain(self) -> None:
             If `pssn_data` is not properly set.
         """
 
-        if self.pssn_data["pssn"] is None or self.pssn_data["sensor_names"] is None:
+        if not self.pssn_data["pssn"] or not self.pssn_data["sensor_names"]:
```
teweitsai, line 365: If `mypy` does not complain, please consider:

```python
if (not self.pssn_data["pssn"]) or (not self.pssn_data["sensor_names"]):
```

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -100,8 +100,14 @@ class OFCData(BaseOFCData):
...
     """
 
-    def __init__(self, name=None, config_dir=None, log=None, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(
+        self,
+        name: str | None = None,
+        config_dir: str | None = None,
+        log: logging.Logger | None = None,
+        **kwargs: object,
```
teweitsai, line 108: For the `**kwargs` or `*args`, please consider the followings and do the related tuning:

https://github.com/lsst-ts/ts_m2gui/blob/develop/python/lsst/ts/m2gui/utils.py#L426-L431

This comment applies to others.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -100,8 +100,14 @@ class OFCData(BaseOFCData):
...
+    ) -> None:
+        super().__init__()
```
teweitsai, line 110: I think the original code is:

```python
super().__init__(**kwargs)
```

It passes the keyword arguments to the parent class.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -604,8 +683,13 @@ def configure_controller(self):
 
             self.xref = self.controller["xref"]
 
-    async def __aenter__(self):
+    async def close(self) -> None:
+        """Clean up any resources held by the OFCData instance."""
+        # Add any necessary cleanup code here
+        pass
```
teweitsai, line 689: Please explain more for this and which ticket would implement the related details. Thanks!

teweitsai (approved):
Looks good.

python/lsst/ts/ofc/ofc_data/ofc_data.py
teweitsai: In this case, you can ignore the mypy check by following:
https://stackoverflow.com/questions/49220022/how-can-mypy-ignore-a-single-line-in-a-source-file

python/lsst/ts/ofc/ofc_data/ofc_data.py
teweitsai: Not sure I understand. I have the `def __aexit__()` here as well but I do not need to create a `async def close()` function to use?
https://github.com/lsst-ts/ts_m2gui/blob/develop/python/lsst/ts/m2gui/controltab/tab_default.py#L207-L215

teweitsai: If it is just a place holder for the future, I would like to ask, will we implement it in the future? If so, please have a ticket number there. If not, why do we need this place holder? I do not understand still...

gmegh: If i don't, then it complains about the following:
```python/lsst/ts/ofc/ofc_data/ofc_data.py:691: error: "OFCData" has no attribute "close"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)```

teweitsai: Please take some time to google people have the similar issue or not. If yes and this is the only way, please put the above comment into the code to let future's people know we need this placeholder. Thanks!

gmegh: Removed the `self.close()` call and I substituted it for a `pass`

PR #64, opened 2024-07-30
Add safety constraints for bending modes

gmegh:


teweitsai:
For the unit of `psi`, I think we should use the SI unit to be consistent with xml.

python/lsst/ts/ofc/bending_mode.py
```diff
@@ -19,41 +19,47 @@
...
 
 import numpy as np
 
 from .ofc_data import OFCData
 from .utils import rot_1d_array
 
 
-class BendModeToForce:
-    """Transform Bend mode to forces.
+class BendingMode:
```
tribeiro, line 30: Why the name change?

tribeiro (changes requested):
Ok, I have one comment that I would like to see addressed before continuing the review.

python/lsst/ts/ofc/bending_mode.py
gmegh: Because BendingModeToForce is not exactly the case anymore. Now we can also go from bendingMode to stresses, so that's why it made more sense conceptually to change the name.

tribeiro (commented):
Please, refactor your changes to only make the changes that are needed for the feature you are planning on adding. These minor naming changes are really distracting.

python/lsst/ts/ofc/bend_mode_to_force.py
```diff
@@ -95,6 +112,30 @@ def __init__(self, component: str, ofc_data: OFCData) -> None:
             :, usecols
         ]
 
+    def stresses(self, dof: np.ndarray[float]) -> np.ndarray[float]:
```
tribeiro, line 115: Consider renaming this method `get_stress_from_dof` or something on these lines.

python/lsst/ts/ofc/bend_mode_to_force.py
```diff
@@ -95,6 +112,30 @@ def __init__(self, component: str, ofc_data: OFCData) -> None:
...
+            Mirror bending mode DOF in um.
+
+        Returns
+        -------
+        `np.ndarray`
+            Mirror stress in psi per bending mode.
+        """
+        # Apply the positive (tensile) or negative (compressive)
+        # bending mode stresses based on the sign of the DOF
+        stresses = np.where(
```
tribeiro, line 130: Is stresses right? Should it be stress instead? I don't know really.

tribeiro (approved):
Looks good! I only have a few minor inline comments. One is about the name of the method you created in `BendModeToForce`. I recommend you expand it a bit to indicate an action being taken (getting the stress from the dofs). Also, I am not sure if the use of the word "stresses" is correct. I have the impression it should be "stress" only, but I am not sure.

tribeiro:
One more comment, I think you should mark these changes for version 4.0.0 instead of 3.3. I would like to get this and all the other major changes you've made recently into one major release. Please, don't release this until we get some of the other PRs and changes you want to make merged in.

PR #68, opened 2024-08-20
Make zn_selected configurable from policy yaml files

gmegh:


policy/configurations/oic_controller.yaml
```diff
@@ -17,6 +17,13 @@ xref: x00
 kp: 1.0
 ki: 0
 kd: 0
+zn_selected: [
+  4, 5, 6, 7, 8,
+  9, 10, 11, 12, 13,
+  14, 15, 16, 17, 18,
+  19, 20, 21, 22, 23, 
+  24, 25, 26, 27, 28
+]
```
teweitsai, line 26: Please add the comment to explain the detail of this new field.

policy/configurations/pid_controller.yaml
```diff
@@ -14,6 +14,13 @@ kp: 0.8
 ki: 0.1
 kd: 0.1
 derivative_filter_coeff: 1.0
+zn_selected: [
```
teweitsai, line 17: Same comment above.

teweitsai (approved):
Looks good! Thanks for the improvement!

PR #70, opened 2024-09-20
Update M2 bending mode actuator forces

gmegh:


policy/M2/M2_1um_72_force.yaml
```diff
@@ -1,727 +1,780 @@
...
-  -8.726511, -220.8554, -31.22918, 6.494293, -7.974032, 128.4726, -23.13773, 180.3884,
-  -308.2483, 4.909395, 40.26356, 271.2906, 32.803, 314.4628, -187.7564, 10.06507,
-  68.11612, 427.9039, -316.9137, 59.20395, 511.1871, 14.9535, -18.98779, -612.2799,
-  -445.3942, 203.5188, -691.1165, 90.43213, 735.4239, -113.4857, 485.2781, 871.6047,
-  83.69699, -153.0873, 7.839825, 1275.462, 285.5154, -458.8959, -755.655, -58.34918,
-  -722.1435, -302.2416, 490.1717, 1095.906, 1163.194, -282.8819, -124.0843, 72.32561,
-  -0.0, 0.0, 0.0]
+---
+
+# September 19, 2024
```
teweitsai, line 3: Please remove the date.

teweitsai (approved):
Looks good.

BTW, you would need to check the coordinate system here as well:

https://github.com/lsst-ts/ts_ofc/blob/develop/policy/M2/rotMatM2.yaml#L10-L11

teweitsai:
You might need to check the sensitivity matrix would need the update or not.

gmegh:
Right, in principle we don't need to change the sensitivity matrix because it was generated using the sagging of the bending modes and not the actuator forces. However, we will wait to merge this PR until @jmeyers314 has verified that the sagging that batoid uses (and therefore, the one the sensitivity matrix used) corresponds to the one Malhar derived from FEA.

jmeyers314:
I think I'd prefer to just multiply the original force matrix by 4.448222 (Newtons per lbf).  I just checked and aside from the difference in overall amplitudes, the force vectors are identical to machine precision ~1e-16.  At some level the detail in the amplitudes don't matter since we'll calibrate these out on sky anyway, so I'd prefer the simpler change.

teweitsai:
In this case, maybe just add a constant such as: `LBF_TO_N = 4.448222` and explain the details in the file that holds the force matrix to make the unit and related transformation clear?

teweitsai:
@jmeyers314 I just realized maybe there was some misunderstanding for this PR. We are asking your help to make sure and left the comment to see the FEA used by Malhar is consistent with the `batoid`.

For the unit of lbs, I think you are talking about another PR related to the stress. I think Guillem had helped to change the unit to SI. Although the unit transformation is not difficult, based on the software maintainability,  it would be easier to keep the same unit. I think the policies files in `ts_wep`, `ts_ofc`, and `ts_mtaos` are basically in SI unit. The `ts_xml` is mainly in SI unit as well.

Thanks!

gmegh:
Hi @teweitsai I think Josh is talking about this PR. The new values of the file I updated basically have transformed the previous values and multiplied them by 4.4 to fix the unit difference. His point is that we should use Bo's original file and multiply by that instead of using the values re-derived.

teweitsai:
I think Bo used the SI unit originally.

gmegh:
As Malhar found out, Bo did not use originally N for this file, hence the 4.4 disagreement.

teweitsai:
When Bo provided the bending mode to `ts_ofc`, I think the unit was in N.

jmeyers314:
Hey @teweitsai ,

  I agree that Bo intended to use and wrote that he used Newtons for the force (e.g. [here](https://github.com/lsst-sitcom/M2_FEA/blob/main/data/M2_1um_72_README.txt#L13-L14)). That seems to be in conflict with Malhar's results though.  Given Bo's note [here](https://github.com/bxin/ZEMAX_FEMAP/blob/main/1M2Bending/README.md?plain=1#L24) on converting from inches to meters, it seems possible that another imperial <-> metric conversion may be in play.  We'll find out for sure when we measure the sensitivity matrix on the sky.

  I also think we all agree that keeping the units consistent between the M2 and M1M3 policy files is desirable.

PR #73, opened 2024-10-04
Fix `get_sensor_names` to work with lsstfam

gmegh:


doc/versionHistory.rst
```diff
@@ -14,6 +14,7 @@ v4.0.0
 * Add function `stresses()` to return calculated stress on the mirrors from bending modes.
 * Add policy files for the bending mode stress per um and scale factor.
 * Add function `set_aggregated_state()` to be able to set the total DOF state.
+* Fix `get_sensor_names` helper to work with lsstfam.
```
suberlak, line 17: Would it not be normally v4.0.1 ?

PR #74, opened 2024-10-16
Filter out NaN zernike values in calculate_corrections

gmegh:


suberlak (approved):
Tested that when provided with a mix of `nan` arrays and valid Zk values we get valid corrections, equal to those that would be obtained if only the sensors with zk estimates were passed

suberlak:
Would you consider updating the test for `ofc.py` that feeds `ofc_calc.calculate_corrections` with a mix of valid measurements and `nan` arrays to test the new functionality?

PR #75, opened 2024-10-23
Fix configuration from obscuration and pupil to pupil and field

gmegh:


policy/configurations/comcam.yaml
```diff
@@ -1,13 +1,13 @@
 description: LSST ComCam Parameters
 instrument: lsst.obs.lsst.ComCam
 
-# Pupil inner and outer radius
-pupil:
+# Field inner and outer radius
+field:
   radius_inner: 0.0 # meter
```
jmeyers314, line 6:
```suggestion
  radius_inner: 0.0 # degrees
```

jmeyers314: And similar below.

jmeyers314 (approved):
One small comment.  Looks good!  Thanks for this.

PR #83, opened 2025-05-13
Tickets/dm 50850

gmegh:


python/lsst/ts/ofc/controllers/zernike_controller.py
```diff
@@ -0,0 +1,356 @@
...
+
+        uk = (
+            self.kp * error
+            + self.ki * self.integral
+            + self.kd * self.filtered_derivative
+        )
+        self.previous_error = error
+        return uk
+
+    def getPsfGradPerZernike(
```
tribeiro, line 283: please, use `snake_case`..

python/lsst/ts/ofc/controllers/zernike_controller.py
```diff
@@ -0,0 +1,356 @@
...
+        num_zk = self.ofc_data.znmax - self.ofc_data.znmin + 1
+        self.zk_state0 = np.zeros((num_sensors, num_zk))
+        self.zk_state = self.zk_state0.copy()
+
+        # Initialize previous error and integral
+        self.previous_error = self.setpoint - self.zk_state0
+        self.integral = self.previous_error
+        self.filtered_derivative = np.zeros((num_sensors, num_zk))
+
+    @property
```
tribeiro, line 110: Why are you using property here? you are just returning `self._kp` and setting it directly in the setter, why not just have `self.kp` as a class attribute directly?

Now, one thing that would be worth doing is have the `kp` property map directly to the `ofc_data` attributes, e.g.:

```
@property
def kp(self) -> float:
    return self.ofc_data.zk_controller["kp"]

@kp.setter
def kp(self, value: float) -> None:
    self.ofc_data.zk_controller["kp"] = value
```

and so on.

python/lsst/ts/ofc/controllers/zernike_controller.py
```diff
@@ -0,0 +1,356 @@
...
+    def kd(self) -> float:
+        return self._kd
+
+    @kd.setter
+    def kd(self, value: float) -> None:
+        self._kd = value
+
+    @property
+    def derivative_filter_coeff(self) -> float:
```
tribeiro, line 135: same as above...

python/lsst/ts/ofc/controllers/zernike_controller.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def set_aggregated_state(self, value: np.ndarray[float]) -> None:
+        """Set the aggregated state.
+
+        Parameters
+        ----------
+        value : `np.ndarray[float]`
+            Aggregated state.
+        """
+        self.zk_state += value
```
tribeiro, line 181: Method is called `set_...` but it is doing the same as `aggregate_...`. Should this be `self.zk_state = value` instead?

python/lsst/ts/ofc/controllers/zernike_controller.py
```diff
@@ -0,0 +1,356 @@
...
+        y = (
+            wfe
+            - get_intrinsic_zernikes(
+                self.ofc_data, filter_name, sensor_names, rotation_angle
+            )
+            - y2_correction
+        )
+        conversion_factors = self.getPsfGradPerZernike(jmax=28)
+
+        zernike_step = self.calculate_pid_step(y * conversion_factors[np.newaxis, :])
```
tribeiro, line 247: I think you should apply this to the wfe only no? instead of using the entire aggregated state?

python/lsst/ts/ofc/ofc.py
```diff
@@ -84,7 +84,11 @@ def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None
 
         self.ofc_data = ofc_data
 
-        self.state_estimator = StateEstimator(self.ofc_data)
+        self.state_estimator = StateEstimator(self.ofc_data, log=self.log)
+
+        self.zernike_controller = ZernikeController(
```
tribeiro, line 89: Is this right? you need the ZernikeController here? Shouldn't this be in `set_controller`?

python/lsst/ts/ofc/ofc.py
```diff
@@ -84,7 +84,11 @@ def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None
...
+            self.ofc_data, num_sensors=4, log=self.log
```
tribeiro, line 90: `num_sensors` is hard-coded to 4...

tribeiro (changes requested):
I made some inline comments. But I am concerned with some of the hard-coding you did here in there.

In some places in the `OFC` and `OFCData` you are hard-coding the new ZernikeController.

PR #112, opened 2026-02-19
[DM-53684] Add initial version of runOfcTask.py

jbkalmbach:


python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,133 @@
...
+
+        if camera.getName() == "LSSTCam":
+            ofc_data = OFCData("lsst")
+        else:
+            raise ValueError(f"Unsupported camera {camera.getName()}")
+
+        ofc_calc = OFC(ofc_data)
+        noll_indices = aggregateZernikesAvg.meta["nollIndices"]
+        ofc_calc.ofc_data.znmin = np.min(noll_indices)
+        ofc_calc.ofc_data.zn_selected = np.arange(np.min(noll_indices), np.max(noll_indices) + 1)
```
gmegh: this should just be equal to noll_indices

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,133 @@
...
+        camera : `lsst.afw.cameraGeom.Camera`
+            Camera object.
+        """
+
+        if camera.getName() == "LSSTCam":
+            ofc_data = OFCData("lsst")
+        else:
+            raise ValueError(f"Unsupported camera {camera.getName()}")
+
+        ofc_calc = OFC(ofc_data)
```
gmegh: you are not setting what dofs you want to use, which means you want to use the default?

gmegh: you probably want to make that a configuration

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,180 @@
...
+        storageClass="Camera",
+        doc="Input camera to construct complete exposures.",
+        dimensions=["instrument"],
+        isCalibration=True,
+        lookupFunction=lookupStaticCalibrations,
+    )
+    ofcCorrections = ct.Output(
+        doc="Visit-level table of OFC corrections",
+        dimensions=("visit", "instrument"),
+        storageClass="AstropyTable",
```
gmegh: why not ArrowAstropy?

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,180 @@
...
+            Camera object.
+
+        Returns
+        -------
+        ofcCorrections : `numpy.ndarray`
+            Corrections for the individual componets. Order is: M2 Hexapod,
+            Camera Hexapod, M1M3 and M2.
+        """
+
+        if camera.getName() == "LSSTCam":
```
gmegh, line 139: why do you need the camera for this? I think you can list instrument in the pipeline connections and that will take care of that, no?

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,180 @@
...
+            raise ValueError(f"Unsupported camera {camera.getName()}")
+
+        ofc_calc = OFC(ofc_data)
+        noll_indices = aggregateZernikesAvg.meta["nollIndices"]
+        j_max = max(noll_indices)
+        j_min = min(noll_indices)
+        ofc_calc.ofc_data.zn_selected = noll_indices
+
+        use_dofs = np.isin(np.arange(50), self.dof_indices)
+        self.log.info(f"Using DOF indices: {np.where(use_dofs)[0]}")
```
gmegh: would be good to also add logging that tells you exactly what dofs are being used not only the indices

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            raise ValueError(f"Unsupported camera {camera.getName()}")
+
+        ofc_calc = OFC(ofc_data)
+        noll_indices = aggregateZernikesAvg.meta["nollIndices"]
+        j_max = max(noll_indices)
+        j_min = min(noll_indices)
+        ofc_calc.ofc_data.zn_selected = noll_indices
+
+        use_dofs = np.isin(np.arange(50), self.dof_indices)
```
gmegh: probably you mean used_dofs?

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,180 @@
...
+
+        ofc_calc.ofc_data.comp_dof_idx = {
+            "m2HexPos": np.array([val for val in use_dofs[:5]], dtype=bool),
+            "camHexPos": np.array([val for val in use_dofs[5:10]], dtype=bool),
+            "M1M3Bend": np.array([val for val in use_dofs[10:30]], dtype=bool),
+            "M2Bend": np.array([val for val in use_dofs[30:]], dtype=bool),
+        }
+        ofc_calc.controller.reset_history()
+
+        self.ofc_calc = ofc_calc
```
gmegh: why wait until here to assign it and not begin by assigning it? Is there any beenfit in doing this? otherwise, it may be cleaner to do it earlier

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,180 @@
...
+
+        wfe_list = list()
+        for wfe in aggregateZernikesAvg[self.column_name]:
+            zern_out = np.zeros(j_max - j_min + 1)
+            for i, noll in enumerate(noll_indices):
+                zern_out[noll - j_min] = wfe[i]
```
gmegh: why not use ts_wep makeDense? Or you just want to avoid the ts_wep dependency? If so, fair enough

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,180 @@
...
+        camera : `lsst.afw.cameraGeom.Camera`
+            Camera object.
+
+        Returns
+        -------
+        `lsst.pipe.base.Struct`
+            A struct containing the visit-level table of OFC corrections.
+        """
+
+        ofcCorrections = self._runOfc(aggregateZernikesAvg, camera)
```
gmegh: as it is, it may make sense to have everything in this function instead of in _runOfc? Unless you plan on adding more things to run

gmegh (commented):
Some additional comments

python/lsst/ts/ofc/task/run_ofc_task.py
jbkalmbach: Yeah, exactly. I didn't want to have to introduce the dependency just for this.

python/lsst/ts/ofc/task/run_ofc_task.py
jbkalmbach: Do we need this at all? It was just easier for testing but I think if we are just running this as a task we don't really need it.

python/lsst/ts/ofc/task/run_ofc_task.py
jbkalmbach: Yeah, I agree. I initially thought it would be more complicated but it ends up being pretty straightforward.

python/lsst/ts/ofc/task/run_ofc_task.py
jbkalmbach: I'll leave it since it's useful if running interactively. (But move it as suggested).

python/lsst/ts/ofc/task/run_ofc_task.py
jbkalmbach: Added a comment in the code about this.

python/lsst/ts/ofc/task/run_ofc_task.py
jbkalmbach: Hmm...maybe but in all our ts_wep tasks we do it this way. Also, requiring something from the connections wouldn't allow a user to run this interactively.

python/lsst/ts/ofc/controllers/pid_controller.py
```diff
@@ -29,12 +32,15 @@
 class PIDController(BaseController):
     """PID controller."""
 
+    def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None:
+        super().__init__(ofc_data, log)
+        self.ofc_data.controller["name"] = "PID"
```
gmegh: Sorry, why are you doing that? That "name" is what tells OFC what controller it should be using... Not the other way around

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -707,6 +707,7 @@ def configure_controller(self) -> None:
             )
 
         if self.controller["name"] not in ["PID", "OIC"]:
+            print(self.controller["name"])
```
gmegh: you probably don't want this here?

python/lsst/ts/ofc/task/run_ofc_task.py
gmegh: Why not? It would, you just need to pass the right instrument when you run this pipeline... Or if you want to run it interactively in a notebook then initialize the instrument to be that at the beginning?

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,192 @@
...
+        j_max = max(noll_indices)
+        j_min = min(noll_indices)
+        self.ofc_calc.ofc_data.zn_selected = noll_indices
+
+        print(self.ofc_calc.ofc_data.controller)
+        self.ofc_calc.set_controller(self.controller_name)
+        if self.truncation_index is not None:
+            self.log.info(f"Setting truncation index to {self.truncation_index}")
+            self.ofc_calc.set_truncation_index(self.truncation_index)
+        print(self.ofc_calc.ofc_data.controller)
```
gmegh: you probably don't want this?

python/lsst/ts/ofc/task/run_ofc_task.py
```diff
@@ -0,0 +1,192 @@
...
+        }
+        self.ofc_calc.controller.reset_history()
+
+        # If we require ts_wep as a prerequisite for functionality
+        # in the future, we can replace this with makeDense from ts_wep.utils.
+        # But for now we avoid the dependency on ts_wep with this.
+        wfe_list = list()
+        for wfe in aggregateZernikesAvg[self.column_name]:
+            zern_out = np.zeros(j_max - j_min + 1)
+            for i, noll in enumerate(noll_indices):
```
gmegh: do you need this for loop, though? can't you just do zern_out[noll_indices - j_min] = wfe?

python/lsst/ts/ofc/utils/utils.py
```diff
@@ -87,3 +88,40 @@ def rot_1d_array(array: np.ndarray[float], rot_mat: np.ndarray[float]) -> np.nda
     rot_array = rot_mat.dot(array2d)
 
     return rot_array.ravel()
+
+
+def get_dof_names(filepath: str, indices: list[int] | None = None) -> list[str]:
```
gmegh: Why have the filepath be passed here? Shouldn't we be able to get these without passing a specific filepath? As in, you could just make this policy file exist in ofc without expecting the user to pass it

python/lsst/ts/ofc/task/run_ofc_task.py
jbkalmbach: Where would you initialize it? There's no butler instance required to run it interactively.

python/lsst/ts/ofc/controllers/pid_controller.py
jbkalmbach: This was due to confusion on the output from pid_controller configuration. Removed.

PR #113, opened 2026-03-16
Make vmodes used constant regardless of rotation angle

gmegh:


jmeyers314 (commented):
This looks right to me.  Have you run it on any real data to compare to previous results?  (Especially near rot=0 where it should be pretty close to the previous way of doing things?)

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -323,12 +323,16 @@ def calculate_pid_step(self, state: np.ndarray[float]) -> np.ndarray[float]:
             either dofs or v-modes.
         """
         error = self.setpoint[self.ofc_data.dof_idx] - state
+        self.log.debug(f"Error before any gains in whichever basis PID is {self.integral}")
```
tribeiro: It would be good if you could merge all these log messages into one. You could do something like:

```
pid_log_message = f"Error before any gains in whichever basis PID is {self.integral}.\n"
...

pid_log_message += f"Integral component before clipping is {self.integral}.\n"

...

pid_log_message += f"Integral component is {self.integral}.\n"

...

pid_log_message += f"Corrections after Kp/Ki/Kd gains is {self.integral}."

self.log.info(pid_log_message)
```

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -165,7 +165,7 @@ def __init__(
         self.start_task.set_result(None)
 
         # Initialize controller configuration
-        self._controller_filename = "oic_controller.yaml"
+        self._controller_filename = "init.yaml"
```
tribeiro, line 168: we really need to make a way to update this through configuration. No need to do anything for now but having this hard coded here with no way to configure it is not great.

Also, remember that the MTAOS reads this from ts_config_mttcs so we will need to update it there as well.

python/lsst/ts/ofc/state_estimator.py
```diff
@@ -83,6 +103,7 @@ def dof_state(
         sensor_names: list,
         rotation_angle: float,
         subtract_intrinsics: bool = True,
+        basis: str = "x",
```
tribeiro: Using a string from something like this is not great. Maybe create an enumeration with the available options and then you can be more explicit about it, e.g.:

```
class ControlBasis(enum.IntEnum):
    DoF = enum.auto()
    VMode = enum.auto()
```

python/lsst/ts/ofc/state_estimator.py
```diff
@@ -282,6 +260,7 @@ def get_sensitivity_matrix(
         normalize: bool = False,
         truncate: bool = False,
         check_invertible: bool = True,
+        basis: str = "x",
```
tribeiro: Same as above to use an enumeration.

python/lsst/ts/ofc/state_estimator.py
```diff
@@ -339,38 +321,37 @@ def get_sensitivity_matrix(
...
-                    self.rcond = 0.99 * s[self.truncate_index - 1] / np.max(s)
-
-            cutoff = self.rcond * np.amax(s, axis=-1, keepdims=True)
-            large = s > cutoff
-            s[~large] = 0
-            sensitivity_matrix = u @ np.diag(s) @ vh
-
-        return sensitivity_matrix
+            if self.truncate_index is None:
+                raise ValueError("Truncation index set incorrectly in the controller.")
```
tribeiro: Might be worth improving this error message to be more descriptive. Something like:

```
raise ValueError(
    "Truncation index set incorrectly in the controller. "
    "Configured to truncate but truncation index is 'None', "
    "check OFC configuration."
)
```

tribeiro (commented):
Here a few comments for you to consider....

gmegh:
@jmeyers314 here is the comparison with previous code and new code. I think they look very similar.

```
DOFs for 2026031500617    w/ vmode corners (rot=0)   w/ vmode dzs           Diff
-----------------------------------------------------------------------------------
M2 dz                      3.65854516                10.79461642         7.13607126
M2 dx                      0.00736509                 0.00171689        -0.00564820
M2 dy                      0.01839665                 0.00412064        -0.01427601
M2 drx    (arcsec)         0.78768682                 0.78767728        -0.00000955
M2 dry    (arcsec)        -0.45614568                -0.45605386         0.00009182
Cam dz                    -1.83752247               -10.83770899        -9.00018652
Cam dx                    -0.00088326                -0.00207350        -0.00119024
Cam dy                    -0.00219430                -0.00512307        -0.00292878
Cam drx   (arcsec)         2.48203317                 2.47424359        -0.00778958
Cam dry   (arcsec)        -1.87326702                -1.86903179         0.00423524
M1M3 B1                   -0.00220638                -0.00563334        -0.00342696
M1M3 B2                    0.01848333                 0.01998282         0.00149949
M1M3 B3                   -0.00981198                -0.02267383        -0.01286185
M1M3 B4                   -0.00585562                -0.00635235        -0.00049673
M1M3 B5                   -0.01003661                -0.01028279        -0.00024618
M1M3 B6                    0.01549561                 0.01531661        -0.00017899
M1M3 B7                    0.03059372                 0.03026306        -0.00033066
M2 B1                     -0.03310112                -0.03170911         0.00139201
M2 B2                      0.02010840                 0.01491238        -0.00519602
M2 B3                     -0.01257931                -0.01214247         0.00043683
M2 B4                      0.00919285                 0.00832774        -0.00086511
M2 B5                     -0.00531953                -0.01614294        -0.01082341
```

tribeiro (approved):
Looks good! thanks for the updates

jmeyers314:
Thanks for running the comparison.  Some interesting differences in the symmetric dofs (dz, M1M3B3, M2B5), but I think we can push forward.

PR #118, opened 2026-04-13
Tickets/rso 249

dsanmartim:


doc/news/RSO-249.feature.rst
tribeiro, line 1: Please, split the different "features" into their own news fragment instead of having them all in a single file like this.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -212,6 +211,11 @@ def __init__(
         self._dof_idx = np.arange(sum([self.comp_dof_idx[comp]["idxLength"] for comp in self.comp_dof_idx]))
         self._dof_idx_mask = np.ones_like(self._dof_idx, dtype=bool)
 
+        # Set the name of the instrument. This reads the instrument-related
```
tribeiro: Do you really need to move this from where it was here? it creates unnecessary diff...

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -322,6 +326,56 @@ def zn_selected(self, value: np.ndarray[int]) -> None:
         self._zn_selected = value
         self._zn_idx_mask = np.isin(self._zn_idx, self._zn_selected - self.znmin)
 
+    @property
+    def vmodes_selected(self) -> np.ndarray[int] | None:
```
tribeiro: One of the benefits of having a property like this is that you can handle the type conversion explicitly and return a single data type. Consider doing something like

```
def vmodes_selected(self) -> np.ndarray[int]:
    """..."""
    return self._vmodes_selected if self._vmodes_selected is not None else self.default_vmodes_selected
```

where `self.default_vmodes_selected` is a selection all all vmodes.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -322,6 +326,56 @@ def zn_selected(self, value: np.ndarray[int]) -> None:
...
+        ValueError
+            If the selected v-modes are invalid for the current DOF mask.
+        """
+        if value is None:
+            self._vmodes_selected = None
+            return
+
+        self._vmodes_selected = self._validate_vmodes_selected(value)
+
+    def _validate_vmodes_selected(
```
tribeiro, line 364: Do you really need to have a separate method to do this? why not do it all inline in the `vmodes_selected` setter? I think it creates unnecessary fragmentation.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -322,6 +326,56 @@ def zn_selected(self, value: np.ndarray[int]) -> None:
...
+        self, value: np.ndarray[int] | list[int], n_vmodes: int | None = None
+    ) -> np.ndarray[int]:
+        """Validate selected v-mode indices against the active DOF mask."""
+        vmodes_selected = np.asarray(value)
+
+        if vmodes_selected.ndim != 1 or vmodes_selected.size == 0:
+            raise ValueError("vmodes_selected must be a one-dimensional, non-empty array.")
```
tribeiro: Consider adding what was wrong with the input here. For example:

```
"vmodes_selected must be a one-dimensional, non-empty array. "
f"Received; {vmodes_selected.ndim=} (expected 1), "
f"{vmodes_selected.size=} (expected >0)."
```

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -322,6 +326,56 @@ def zn_selected(self, value: np.ndarray[int]) -> None:
...
+
+        if not np.all(np.equal(vmodes_selected, vmodes_selected.astype(int))):
+            raise ValueError("V-mode indices must be integers.")
```
tribeiro: You could cast these as `int` and avoid this issue no? For example, in line 360 you could do:
 
```
vmodes_selected = np.array(np.asarray(value), dtype=int)
```

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -322,6 +326,56 @@ def zn_selected(self, value: np.ndarray[int]) -> None:
...
+
+        vmodes_selected = np.sort(vmodes_selected.astype(int))
```
tribeiro: I think you could also do this in one step in line 360.

```
vmodes_selected = np.sort(np.array(np.asarray(value), dtype=int))
```

you might even trow in a "unique" to make it even more robust:

```
vmodes_selected = np.sort(np.unique(np.array(np.asarray(value), dtype=int)))
```

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -322,6 +326,56 @@ def zn_selected(self, value: np.ndarray[int]) -> None:
...
+        max_vmode = len(self.dof_idx) if n_vmodes is None else n_vmodes
+
+        if np.any(vmodes_selected < 1) or np.any(vmodes_selected > max_vmode):
+            raise ValueError(f"V-mode index must be between 1 and {max_vmode}.")
```
tribeiro: Add the received value so people know what was wrong, e.g.;

```
all_vmodes_selected = ",".join(vmodes_selected)
mask = np.bitwise_or(
vmodes_selected < 1,
vmodes_selected > max_vmode,
)
bad_vmodes_selected = ",".join(vmodes_selected[mask])

raise ValueError(
    f"V-mode index must be between 1 and {max_vmode}. "
    f"Received vmodes ({all_vmodes_selected}) contains the following invalid values; {bad_vmodes_selected}."
)

```

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -322,6 +326,56 @@ def zn_selected(self, value: np.ndarray[int]) -> None:
...
+
+        if len(np.unique(vmodes_selected)) != len(vmodes_selected):
+            raise ValueError("V-mode indices must be unique.")
```
tribeiro: If you apply the unique selection you can remove this....

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -714,6 +776,11 @@ def configure_controller(self) -> None:
         if "zn_selected" in self.controller:
             self.zn_selected = np.array(self.controller["zn_selected"])
 
+        if "vmodes_selected" in self.controller:
```
tribeiro: This can be simplified with:

```
self.vmodes_selected = get(self.controller, "zn_selected", None)
```

python/lsst/ts/ofc/state_estimator.py
```diff
@@ -97,6 +97,62 @@ def _update_from_ofc_data(self, ofc_data: OFCData) -> None:
...
+            f"with singular values {self.S[self.effective_vmode_idx].tolist()}."
+        )
+
+    def _compute_effective_vmode_indices(self) -> np.ndarray[int]:
+        """Compute the effective v-mode indices from config.
+
+        If ``vmodes_selected`` is set explicitly in ``ofc_data``, those
+        indices are used (converted from 1-based to 0-based). Otherwise,
+        automatic selection is performed using either ``truncation_index``
+        or ``truncation_threshold`` (rcond).
```
tribeiro: Please, document the return value.

tests/test_ofc.py
```diff
@@ -161,6 +161,46 @@ def test_calculate_corrections(self) -> None:
...
+            sensor_ids=self.sensor_id_list,
+            filter_name=filter_name,
+            rotation_angle=rotation_angle,
+            subtract_intrinsics=True,
+            control_vmodes=False,
+        )
+
+        # Check that the last-visit DOF has changed, indicating that the
+        # selected v-modes affected the correction calculation.
+        self.assertFalse(np.allclose(self.ofc.lv_dof, baseline_lv_dof))
```
tribeiro: I think instead of doing this, which is easy to be falsified, it would be better to check against a result we know should be true. For example, you could check which v-modes are excited by the `wfe` and than try removing those v-modes and checking that the dof for those won't change...

tribeiro (approved):
Looks good, I have a few inline comments but they are mostly minor comments with style suggestions.

python/lsst/ts/ofc/ofc_data/ofc_data.py
dsanmartim: Kind of... Note that the `self.name = name` triggers configure_controller(), which now reads `self.vmodes_selected`. The attribute must exist before that call. We could alternatively keep `self.name` in its original position and just add `self._vmodes_selected = None` before it, if you prefer the smaller diff.

dsanmartim: No!

python/lsst/ts/ofc/ofc_data/ofc_data.py
dsanmartim: I think you mean this, right?

`self.vmodes_selected = self.controller.get("vmodes_selected", None)`

python/lsst/ts/ofc/ofc_data/ofc_data.py
dsanmartim: I think we do need it.  The `_validate_vmodes_selected` method is also called in the `comp_dof_idx` setter, since it must validate the existing `vmodes_selected` against the new DOF count before committing the mask change. For example, `vmodes_selected = [1, 2, ..., 20]` is valid with 22 active DOFs but invalid if the new mask reduces to 5 DOFs. The validation must run with a different `n_vmodes` than the setter uses, which is why the `n_vmodes` parameter exists. So, to avoid duplication I think it is better to keep it.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -322,6 +326,88 @@ def zn_selected(self, value: np.ndarray[int]) -> None:
...
+        vmodes_selected : `np.ndarray` [`int`]
+            Sorted, validated array of v-mode indices.
+
+        Raises
+        ------
+        ValueError
+            If input is not a 1-d non-empty array, or contains
+            out-of-range values.
+        """
+        vmodes_selected = np.asarray(value)
```
gmegh, line 388: isn't this an array already at this point?

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -372,7 +460,13 @@ def comp_dof_idx(self, value: dict) -> None:
                 )
             if not isinstance(value[comp], np.ndarray) or value[comp].dtype.type is not np.bool_:
                 raise RuntimeError("Input should be np.ndarray of type bool.")
-            self._dof_idx_mask[start_idx:end_idx] = value[comp]
+            new_dof_idx_mask[start_idx:end_idx] = value[comp]
+
+        if self._vmodes_selected is not None:
+            n_vmodes = int(np.count_nonzero(new_dof_idx_mask))
+            self._vmodes_selected = self._validate_vmodes_selected(self._vmodes_selected, n_vmodes=n_vmodes)
```
gmegh, line 467: Why do you need to run this here, if the setter also runs this function later in L811?

python/lsst/ts/ofc/state_estimator.py
```diff
@@ -97,6 +97,67 @@ def _update_from_ofc_data(self, ofc_data: OFCData) -> None:
...
+        n_vmodes = len(self.ofc_data.dof_idx)
+
+        if self.truncate_index is not None:
+            if self.truncate_index > n_vmodes:
+                self.log.warning(
+                    f"Truncation index ({self.truncate_index}) exceeds the number of used DOF ({n_vmodes}). "
+                )
+            return np.arange(min(self.truncate_index, n_vmodes), dtype=int)
+
+        if self.rcond is not None:
```
gmegh, line 141: what if both truncate_index and rcond are set?

python/lsst/ts/ofc/state_estimator.py
```diff
@@ -213,14 +272,23 @@ def get_vmodes_from_dofs(self, dof: np.ndarray[float]) -> np.ndarray[float]:
...
+            Optical state in the basis of v-modes. Returns a vector with the
+            same dimension as the number of used DOF, with unselected modes
+            padded with zeros.
         """
-        return (
-            np.linalg.inv(self.normalization_matrix)
-            @ dof[self.ofc_data.dof_idx]
-            @ self.Vh[: self.truncate_index].T
-        )
+        dof = np.asarray(dof)
```
gmegh: isn't this already an array given the type constraint?

python/lsst/ts/ofc/state_estimator.py
gmegh: i think you need to update the docstring

python/lsst/ts/ofc/state_estimator.py
```diff
@@ -97,6 +97,67 @@ def _update_from_ofc_data(self, ofc_data: OFCData) -> None:
...
+
+        raise ValueError(
+            "Either truncation_threshold or truncation_index must be configured "
+            "when vmodes_selected is not set."
+        )
+
+    def _pad_vmode_vector(self, v_modes: np.ndarray[float]) -> np.ndarray[float]:
+        """Pad a compact v-mode vector to the full active DOF dimension."""
+        padded_v_modes = np.zeros(len(self.ofc_data.dof_idx))
+        padded_v_modes[self.effective_vmode_idx] = np.asarray(v_modes)
```
gmegh: i don't think you need np.asarray

gmegh (changes requested):
Looks good. Thank you. I left some comments, I would appreciate it if you could address them.

tests/test_ofc.py
dsanmartim: I tested that removing the dominant v-mode changes corrections, but couldn't reliably test that removing an inactive mode leaves corrections unchanged. It seems that removing any v-mode,  even one with near zero singular value -- changes the least-squares subspace, causing the solution to redistribute across remaining modes.

python/lsst/ts/ofc/ofc_data/ofc_data.py
dsanmartim: Not necessarily. The method accepts list[int] as well (from config input), so np.asarray is needed for the .ndim and .size checks.

python/lsst/ts/ofc/state_estimator.py
dsanmartim: Setting both is prevented by `configure_controller()` in `ofc_data.py` which raises `ValueError` if both `truncation_index` and `truncation_threshold` are set. But if they somehow coexist, `truncate_index` wins. I have updated the docstring with it.

python/lsst/ts/ofc/ofc_data/ofc_data.py
dsanmartim: The `comp_dof_idx` setter validates `vmodes_selected` against the new DOF count before committing the mask change. For example, vmodes_selected = [1, 2, ..., 20] is valid with 22 active DOFs but invalid if the new mask reduces to 10 DOFs. The setter at L811 validates against the current DOF count. They use different `n_vmodes` values, which is why the separate validation is needed here.

tests/test_ofc.py
gmegh: how much do they change? they should be small changes, no?

gmegh (approved):
LGTM thanks

PR #119, opened 2026-04-15
Update generate_normalization_weights.py

gmegh:


theoschutt (approved):
Looks good to me. Tested that both the `legacy` and `quadrature` options work. My only suggestion would be to add the following block to make it easier to run as a standalone script (I don't think it conflicts with a different usage of the script?).

```
if __name__ == "__main__":
    generate_normalization_weights()
```

PR #121, opened 2026-05-20
Adding leaky integral code and code to limit the number of terms in t…

craiglagegit:
In the current PID software, the integral terms adds all corrections from the beginning of the run.  This will add two changes:
(1) Implement a maximum number of terms to be included in the integral term.
(2) Implement a “leaky integrator”, where older terms in the integral are deweighted by a constant factor.

In simulations, these changes have been shown to reduce oscillation due to including the Ki term.


craiglagegit:
This is just a first pass.  I'm sure there are things missing or conventions I haven't followed, but I think I have captured the essence of the changes.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -63,12 +63,16 @@ class BaseController:
         Logger class used for logging operations.
     ofc_data : `OFCData`
         OFC data.
-    previous_error : `np.array`
-        Previous error.
+    previous_error : `list`
```
tribeiro: Instead of a list, you might want to consider using a `collections.deque` with `max_len=n_items`

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -63,12 +63,16 @@ class BaseController:
...
+        List of previous errors.
     pssn_data : `dict`
         PSSN data.
     setpoint : `np.array`
         Setpoint for the PID controller.
+    n_iterms : `int`
```
tribeiro: I don't see this attribute defined anywhere....

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -63,12 +63,16 @@ class BaseController:
...
+        Number of terms to be included in the integral.
+    i_factor : `float`
```
tribeiro: I also don't see this attribute defined anywhere.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -92,7 +96,9 @@ def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None
         self._kd = self.format_gain_array(self.ofc_data.controller["kd"], label="kd")
         self._derivative_filter_coeff = self.ofc_data.controller["derivative_filter_coeff"]
         self.setpoint = np.array(self.ofc_data.controller["setpoint"])
-
+        self._n_iterms = self.ofc_data.controller["n_iterms"]
```
tribeiro: I think you need to update the controller configuration to add this new parameter (same for i_factor).

tribeiro (commented):
I think there are a few issues with the implementation you might be able to catch by running the unit tests... But it seems like a sound idea....

python/lsst/ts/ofc/controllers/base_controller.py
craiglagegit: I've made this change.

python/lsst/ts/ofc/controllers/base_controller.py
craiglagegit: I've added the two new terms to the three yaml files.  I'm now able to import it in a notebook with no fails.

craiglagegit:
I've made the changes @tribeiro requested.  I also simplified the integral calculation so that it is a simple array multiplication and that calculate_integral function is no longer required.  It imports in a notebook, but I don't understand the errors being generated.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -324,31 +340,32 @@ def calculate_pid_step(self, state: np.ndarray[float]) -> np.ndarray[float]:
...
             self.derivative_filter_coeff * derivative
             + (1 - self.derivative_filter_coeff) * self.filtered_derivative
         )
-
         uk = (
             self.kp[self.ofc_data.dof_idx] * error
             + self.ki[self.ofc_data.dof_idx] * self.integral
             + self.kd[self.ofc_data.dof_idx] * self.filtered_derivative
         )
+        print(uk.shape)
```
gmegh: this should not be here, no?

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -92,7 +98,9 @@ def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None
         self._kd = self.format_gain_array(self.ofc_data.controller["kd"], label="kd")
         self._derivative_filter_coeff = self.ofc_data.controller["derivative_filter_coeff"]
         self.setpoint = np.array(self.ofc_data.controller["setpoint"])
-
+        self._n_iterms = self.ofc_data.controller["n_iterms"]
+        self._i_factor = self.ofc_data.controller["i_factor"]
```
gmegh: you either need to define a default value in case this is not present in the controller or need to modify ofc_data to ensure any controller configuration has this attribute

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -92,7 +98,9 @@ def __init__(self, ofc_data: OFCData, log: logging.Logger | None = None) -> None
         self._kd = self.format_gain_array(self.ofc_data.controller["kd"], label="kd")
         self._derivative_filter_coeff = self.ofc_data.controller["derivative_filter_coeff"]
         self.setpoint = np.array(self.ofc_data.controller["setpoint"])
-
+        self._n_iterms = self.ofc_data.controller["n_iterms"]
```
gmegh: why do you have a subscript in front of it? I have seen that in the past used whenever we have a setter, but we don't have any here, no?

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -324,31 +340,32 @@ def calculate_pid_step(self, state: np.ndarray[float]) -> np.ndarray[float]:
         """
         error = self.setpoint[self.ofc_data.dof_idx] - state
         pid_log_message = f"Error before any gains in whichever basis PID is {error}.\n"
-
-        self.integral += error
+        if self.integral_weights.shape[1] != len(self.ofc_data.dof_idx):
```
gmegh: When would this ever happen?

gmegh (commented):
Some comments

python/lsst/ts/ofc/controllers/base_controller.py
craiglagegit: Yes.  I'll remove that.

python/lsst/ts/ofc/controllers/base_controller.py
craiglagegit: What I was doing was modifying ts_config_mttcs to have those terms.  That seems to work.  Is that sufficient?

python/lsst/ts/ofc/controllers/base_controller.py
craiglagegit: The other terms like kp, ki, derivative_filter_coeff had the underscore, so I followed suit.  I don't really understand what it does.  Should I remove the underscore?

python/lsst/ts/ofc/controllers/base_controller.py
craiglagegit: When it is defined in the init function, self.ofc_data.dof_idx = 50, so it is orignially defined with 50 terms.  I guess self.ofc_data.dof_idx gets redefined later.

python/lsst/ts/ofc/controllers/base_controller.py
gmegh: no, you'll need to add it in ofc_data.py, see what it's done there for any new attribute in ofc_controller

python/lsst/ts/ofc/controllers/base_controller.py
gmegh: I think that's why we have the reset function, no?

python/lsst/ts/ofc/controllers/base_controller.py
gmegh: The reason is that those are properties. See here https://github.com/lsst-ts/ts_ofc/pull/121/changes#diff-00eee172e0e4c1bd54afed29d9270f21e217d9d1c0b9fd1c5e77a22c855cd7b9R237-R246

gmegh:
@dsanmartim may I ask you to please help @craiglagegit in getting this PR through the finish line? I think it will be very easy for you to figure out the remaining details that Craig is unfamiliar with. Thank you very much!

craiglagegit:
Hi @dsanmartim .  As a check, I imported the latest code into a notebook after your changes, and ran my test run.  It looks identical to what it was before your changes.  Thanks for your help!

dsanmartim:
Hi folks. I believe I have addressed Guillem's comments. You will also notice that I added a new keyword to the PID controller called `use_leaky_integrator` (with retro-compatibility), which can be used to enable and disable the leaky integrator feature. After Guillem takes a final look at the updates I made, it would be good if @craiglagegit can test the updated code in his notebook.

gmegh (approved):
LGTM. Please test on BTS

craiglagegit:
Hi @dsanmartim .  I have tested the updated code in my notebook, and it behaves the same.  Since the default for use_leaky_integrator is True, I didn't have to change anything.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -234,6 +252,15 @@ def derivative_filter_coeff(self) -> float:
     def derivative_filter_coeff(self, value: float) -> None:
         self._derivative_filter_coeff = value
 
+    @property
```
tribeiro: Why use a property and a setter? it doesn't seem like you are doing anything special when setting or retrieving the values from `self._use_leaky_integrator` to worth the overhead. Just set `self.use_leaky_integrator` as a variable would give you the same result.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -277,7 +304,8 @@ def reset_history(self) -> None:
         """Reset the history of the controller."""
         self.dof_state0 = self.dof_state.copy()
         self.integral = np.zeros(len(self.ofc_data.dof_idx))
-        self.previous_error = np.zeros(len(self.ofc_data.dof_idx))
+        for i in range(self._n_iterms):
```
tribeiro: I think what you really want to do here is:

```
self.previous_error: deque[np.ndarray] = deque(maxlen=self._n_iterms)
```

and not fill it up with zeros.

python/lsst/ts/ofc/controllers/base_controller.py
```diff
@@ -324,31 +352,34 @@ def calculate_pid_step(self, state: np.ndarray[float]) -> np.ndarray[float]:
         """
         error = self.setpoint[self.ofc_data.dof_idx] - state
         pid_log_message = f"Error before any gains in whichever basis PID is {error}.\n"
-
-        self.integral += error
+        if self.use_leaky_integrator:
+            if self.integral_weights.shape[1] != len(self.ofc_data.dof_idx):
+                self.integral_weights = self.integral_weights[:, self.ofc_data.dof_idx]
+            self.integral = np.sum(self.integral_weights * np.array(self.previous_error), axis=0)
```
tribeiro: I think this is ignoring the current `error`, no? You might want to move line 382 to before the if statement.

tribeiro (changes requested):

python/lsst/ts/ofc/controllers/base_controller.py
dsanmartim: Had to do this instead: `self.previous_error = deque(maxlen=self._n_iterms)`... mypy complains the type was already defined in the `__init__`.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -86,10 +86,16 @@ class OFCData(BaseOFCData):
     name : `string`
         Name of the instrument configuration. This is used to define where
         `intrinsic_zk` and `y2` will be read from.
+    n_iterms : `int`
```
tribeiro: This is a property of the class, should not be documented as an attribute.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -86,10 +86,16 @@ class OFCData(BaseOFCData):
...
+        Number of error terms in the leaky integrator.
     sensitivity_matrix : `np.ndarray` of `float`
         Sensitivity matrix M.
     start_task : `asyncio.Future`
         Asyncio future that tracks whether the class is setup and ready or not.
+    i_factor : `float`
```
tribeiro: This is a property of the class, should not be documented as an attribute.

python/lsst/ts/ofc/ofc_data/ofc_data.py
```diff
@@ -86,10 +86,16 @@ class OFCData(BaseOFCData):
...
+        Decay factor for older error terms in the leaky integrator.
+    use_leaky_integrator : `bool`
```
tribeiro: This is a property of the class, should not be documented as an attribute.

python/lsst/ts/ofc/controllers/base_controller.py
dsanmartim: You are right. These were documented as public `BaseController` attributes, but they are not public attributes of this class.

The leaky-integrator configuration now lives in `OFCData` as config-backed properties. `BaseController` reads those values during initialization and stores only the internal values it needs as `_n_iterms` and `_i_factor`.

I removed `n_iterms` and `i_factor` from the `BaseController` Attributes section.

python/lsst/ts/ofc/controllers/base_controller.py
dsanmartim: Same as above.

python/lsst/ts/ofc/ofc_data/ofc_data.py
dsanmartim: Good point. These are properties, so they should not be listed in the class Attributes section.

I removed `n_iterms`, `i_factor`, and `use_leaky_integrator` from the `OFCData` Attributes section. They remain documented in their corresponding property methods.

python/lsst/ts/ofc/ofc_data/ofc_data.py
dsanmartim: Same as above.

python/lsst/ts/ofc/ofc_data/ofc_data.py
dsanmartim: Same as above.

